PR #135, opened 2026-03-28
feat(messages): add --since flag to thread command

### LINKED ISSUE #127: feat: add --since flag to messages thread command
## Problem

`slck messages thread <channel> <thread-ts>` always returns the full thread (up to `--limit`). When polling a thread for new messages, the caller has to fetch all messages every cycle and filter client-side.

## Proposal

Add a `--since <timestamp>` flag that only returns messages with `ts` greater than the given value.

```bash
slck messages thread C12345 1700000000.000000 --since 1700000050.000000 -o json
```

This maps to Slack's `conversations.replies` API parameter `oldest`.

## Use case

Polling a thread for new messages at a regular interval (e.g. every 2s). Currently fetches up to 50 messages per poll and deduplicates client-side. With `--since`, the API call itself returns only new messages — less data, less filtering.

Not urgent — current approach works fine for typical thread sizes. Nice-to-have for efficiency.

rianjs:
## Summary
- Adds `--since <timestamp>` flag to `slck messages thread` that maps to Slack's `oldest` parameter on `conversations.replies`
- Only returns messages with `ts` greater than the given value — useful for polling threads without refetching all history
- Consistent with existing `--oldest` flag on `messages history`

Closes #127

## Test plan
- [x] `TestClient_GetThreadReplies_WithOldest` — verifies `oldest` param is sent to API
- [x] Existing `GetThreadReplies` tests updated and passing
- [x] `make lint` clean

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -586,7 +586,7 @@ func (c *Client) GetChannelHistory(channel string, limit int, oldest, latest str
 }
 
 // GetThreadReplies returns replies to a thread (handles pagination to reach requested limit)
-func (c *Client) GetThreadReplies(channel, threadTS string, limit int) ([]Message, error) {
+func (c *Client) GetThreadReplies(channel, threadTS string, limit int, oldest string) ([]Message, error) {
 	var allMessages []Message
 	cursor := ""
 	remaining := limit
@@ -601,6 +601,9 @@ func (c *Client) GetThreadReplies(channel, threadTS string, limit int) ([]Messag
 			batchSize = 200
 		}
 		params.Set("limit", fmt.Sprintf("%d", batchSize))
+		if oldest != "" {
+			params.Set("oldest", oldest)
+		}
 		if cursor != "" {
 			params.Set("cursor", cursor)
 		}
--- a/internal/client/client_test.go
+++ b/internal/client/client_test.go
@@ -473,13 +473,42 @@ func TestClient_GetThreadReplies_Success(t *testing.T) {
 	defer server.Close()
 
 	client := NewWithConfig(server.URL, "test-token", nil)
-	messages, err := client.GetThreadReplies("C123", "1234567890.123456", 100)
+	messages, err := client.GetThreadReplies("C123", "1234567890.123456", 100, "")
 
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
 	if len(messages) != 2 {
 		t.Errorf("expected 2 messages, got %d", len(messages))
+	}
+}
+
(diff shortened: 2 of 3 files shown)


rianjs:
## TDD Assessment: `--since` flag for `thread` command

### What changed
- `GetThreadReplies` gained an `oldest string` parameter
- `thread.go` gained a `--since` flag wired to `opts.since`, which is passed directly to `GetThreadReplies`
- `oldest` is conditionally set in the API params only when non-empty

---

### Client layer (`client_test.go`) — adequate

| Scenario | Covered? |
|---|---|
| `oldest` sent correctly when non-empty | Yes — `TestClient_GetThreadReplies_WithOldest` asserts `oldest=1700000050.000000` via the server handler |
| Empty `oldest` not sent | Partial — `TestClient_GetThreadReplies_Success` passes `""` but never asserts the `oldest` query param is absent from the request |
| Pagination with `oldest` set | Not tested — only a single page (empty cursor) is exercised in `WithOldest` |

The missing "empty oldest not sent" assertion is a minor gap. The current test would pass even if the code unconditionally sent `oldest=`, because the handler only checks that `oldest == "1700000050.000000"` in the positive case, not that it is absent in the base case.

---

### Command layer (`messages_test.go`) — gap

None of the three existing `TestRunThread_*` tests exercise the `--since` flag at all. `opts.since` is left at its zero value (`""`) in every case. There is no test that:

1. Passes a non-empty `since` value and verifies `oldest` reaches the API call.
2. Verifies the flag default is `""` (i.e. omitted from the request by default).

The command layer is the integration point between flag parsing and the client call. Without a test here, a regression that strips or ignores `opts.since` before calling `GetThreadReplies` would go undetected.

---

### Other observations

- **No validation of the `--since` timestamp format.** The flag is a raw string passed directly to the API. An obviously invalid value (e.g. `--since foo`) will be forwarded without error. The history command (`history.go`) has comparable behaviour, so this is consistent, but worth a note if strict UX is desired.
- **`since` vs `oldest` naming mismatch** — the flag is `--since` but the underlying parameter is `oldest`. This is fine UX-wise, but a comment in the code or a test name making the mapping explicit would help future readers.

---

### Summary

Coverage is **adequate at the client layer** for the happy path, but has two concrete gaps:

1. **Client test:** No assertion that `oldest` is absent when `""` is passed.
2. **Command test:** No `runThread` test exercises `opts.since != ""` end-to-end.

Recommend adding one `TestRunThread_WithSince` test in `messages_test.go` that sets `opts.since` and asserts the `oldest` query param reaches the mock server, and tightening `TestClient_GetThreadReplies_Success` to assert `r.URL.Query().Get("oldest") == ""`.

PR #142, opened 2026-04-19
fix(messages): render file attachments and rich_text blocks in text output

### LINKED ISSUE #141: Thread/history text output drops file attachments and rich-text block content
## Problem

The text output of `slck messages thread` and `slck messages history` silently omits two kinds of content that exist in the underlying Slack API response. Both make the default text output unreliable for humans and agents that consume it — they see a thread that looks incomplete when it isn't.

## How this surfaced

A long real-world thread had a bot ("unnamed-hero") responding repeatedly with things like:

> It looks like the files didn't come through — I don't see any attachments in your message.

> I'm not receiving it — the message that came through to me was empty.

The bot was reading thread content via `slck messages thread`. The Slack API response was fine. The `slck` text renderer was what dropped the content.

Reproducer thread: https://signalft.slack.com/archives/C06TAECLME3/p1776363343792319

## Bug 1 — file attachments are stripped from text output

Affected files:
- `internal/cmd/messages/thread.go:67-76`
- `internal/cmd/messages/history.go:65-74`

The `Message` struct (`internal/client/client.go:253`) already has a populated `Files []File` field when a message has uploads. That field flows through to `slck messages thread --output json` correctly. But both text-rendering loops only print `m.Text` and ignore `m.Files` entirely — so a message like "Here's the two check files" renders with no indication that two CSV files were attached.

A user or agent reading the text output has no signal that attachments exist, and no way to know the file IDs needed to invoke `slck files download <id>`.

## Bug 2 — rich-text blocks are dropped entirely

Affected file: `internal/client/client.go:253` (the `Message` struct).

When Slack messages contain rich content — the most common cause is a user pasting a formatted table with CMD+V — the API returns the content in a `blocks` array with `"text": ""`. Our `Message` struct has no `Blocks` field, so JSON unmarshaling silently discards the data, and the text renderer prints an empty line.

The reproducer thread has two such messages: users pasted tables from a PDF, the text-output showed nothing, and the bot (and any human reading the output) had no idea content was there.

## Fix scope

A comprehensive implementation plan will be posted as a comment on this issue. Tracking both bugs in a single issue because they share a root cause (the text renderer only looks at `m.Text`) and will ship together in one PR.

rianjs:
## Summary

- Render file attachments as tab-indented `[file] <name> (<type>, <size>) — slck files download <id>` lines after the message in `slck messages thread` and `slck messages history`. Previously `m.Files` was populated but the text renderer dropped it entirely, leaving readers (humans or agents) with no hint that attachments existed.
- Render Slack rich-text blocks that come through with empty `m.Text` (common when a user pastes a formatted table with CMD+V). A new `Block`/`RichTextElement` model plus `RenderBlocks` walker handles all rich-text sub-types (sections, lists with honored `offset`, preformatted, quotes) and inline element types (text with bold/italic/code/strike, link, user, channel, emoji, broadcast, usergroup, date). Unknown types silently drop so `m.Text` fallback wins when rendering would yield nothing useful.
- Thread preserves newlines from block-rendered bodies and indents continuation lines under the `[<ts>] <user>:` header. Plain-text messages keep their existing single-line rendering — no regression for the common path. History's compact view still truncates to 80 chars.
- `UserResolver.Resolve` and `ResolveMentions` are now nil-safe so `RenderBlocks` can be called with a nil resolver without panicking.

## Test plan

- [x] `make test` — 588 passed across 18 packages (36 new)
- [x] `make lint` — 0 issues
- [x] Manual verification against the reproducer thread: previously-blank table-paste messages now render their content; file-upload messages show `[file] ... slck files download <id>` hints.
- [x] `--output json` still exposes the `files` array and now also exposes `blocks` unchanged.

Closes #141

### DIFF
--- a/internal/client/blocks.go
+++ b/internal/client/blocks.go
@@ -1,0 +1,242 @@
+package client
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// Block is a top-level Slack Block Kit block. Only rich_text is rendered
+// today; other types (section, image, context, etc.) are preserved in the
+// JSON representation but yield empty output from RenderBlocks, letting
+// callers fall back to the message's m.Text plain-text fallback.
+type Block struct {
+	Type     string            `json:"type"`
+	BlockID  string            `json:"block_id,omitempty"`
+	Elements []json.RawMessage `json:"elements,omitempty"`
+}
+
+// RichTextElement is one element inside a rich_text block. The zero value
+// plus the non-nil Elements signal a container (rich_text_section,
+// rich_text_list, rich_text_preformatted, rich_text_quote). Otherwise the
+// element is an inline primitive (text, link, user, channel, emoji,
+// broadcast, usergroup, date).
+type RichTextElement struct {
+	Type     string            `json:"type"`
+	Elements []json.RawMessage `json:"elements,omitempty"`
+
+	// rich_text_list fields (only populated when Type == "rich_text_list")
+	ListStyle string `json:"-"` // "bullet" | "ordered" — populated from polymorphic "style"
+	Indent    int    `json:"indent,omitempty"`
+	Offset    int    `json:"offset,omitempty"`
+	Border    int    `json:"border,omitempty"`
+
+	// inline element fields
+	Text        string `json:"text,omitempty"`
+	URL         string `json:"url,omitempty"`
+	UserID      string `json:"user_id,omitempty"`
+	ChannelID   string `json:"channel_id,omitempty"`
+	Name        string `json:"name,omitempty"`      // emoji name
+	Range       string `json:"range,omitempty"`     // broadcast: here|channel|everyone
+	Timestamp   int64  `json:"timestamp,omitempty"` // date element (unix seconds)
+	Format      string `json:"format,omitempty"`    // date format (not interpreted — Fallback/RFC3339 used instead)
+	Fallback    string `json:"fallback,omitempty"`  // date fallback text
+	UsergroupID string `json:"usergroup_id,omitempty"`
+
+	TextStyle *RichTextStyle `json:"-"` // populated from polymorphic "style"
+}
+
+// RichTextStyle captures inline text formatting flags on a "text" element.
+type RichTextStyle struct {
+	Bold   bool `json:"bold,omitempty"`
+	Italic bool `json:"italic,omitempty"`
+	Code   bool `json:"code,omitempty"`
+	Strike bool `json:"strike,omitempty"`
+}
+
+// UnmarshalJSON handles the "style" field's polymorphism: rich_text_list
+// uses a string ("bullet"|"ordered"), inline text uses an object of
+// boolean flags. Everything else is unmarshaled via the standard rules
+// after stripping "style" out to avoid the default unmarshaler choking.
+func (e *RichTextElement) UnmarshalJSON(data []byte) error {
+	type alias RichTextElement
+	aux := struct {
+		*alias
+		Style json.RawMessage `json:"style,omitempty"`
+	}{alias: (*alias)(e)}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+	if len(aux.Style) == 0 {
+		return nil
+	}
+	trimmed := strings.TrimLeft(string(aux.Style), " \t\n\r")
+	if strings.HasPrefix(trimmed, `"`) {
+		var s string
+		if err := json.Unmarshal(aux.Style, &s); err != nil {
+			return err
+		}
+		e.ListStyle = s
+		return nil
+	}
+	if strings.HasPrefix(trimmed, "{") {
+		var ts RichTextStyle
+		if err := json.Unmarshal(aux.Style, &ts); err != nil {
+			return err
+		}
+		e.TextStyle = &ts
+	}
+	return nil
+}
+
+// MarshalJSON round-trips the polymorphic style field so --output json
+// preserves whichever variant was on the wire.
+func (e RichTextElement) MarshalJSON() ([]byte, error) {
+	type alias RichTextElement
+	aux := struct {
+		alias
+		Style interface{} `json:"style,omitempty"`
+	}{alias: alias(e)}
+	switch {
+	case e.ListStyle != "":
+		aux.Style = e.ListStyle
+	case e.TextStyle != nil:
+		aux.Style = e.TextStyle
+	}
+	return json.Marshal(aux)
+}
+
+// RenderBlocks walks top-level blocks and returns plain text suitable for
+// slck's text output. Returns "" if blocks is empty or no block yields
+// output — callers can use that signal to fall back to m.Text. Safe to
+// call with a nil resolver.
+func RenderBlocks(blocks []Block, resolver *UserResolver) string {
+	if len(blocks) == 0 {
+		return ""
+	}
+	var out strings.Builder
+	for _, b := range blocks {
+		if b.Type != "rich_text" {
+			// Non-rich_text blocks (section, image, context, etc.) are
+			// dropped so the caller falls back to m.Text, which Slack
+			// populates as a plain-text equivalent for these cases.
+			continue
+		}
+		out.WriteString(renderRichText(b.Elements, resolver, 0))
+	}
+	return strings.TrimRight(out.String(), "\n")
+}
+
+// renderRichText walks the sub-elements of a rich_text block.
+func renderRichText(raws []json.RawMessage, resolver *UserResolver, baseIndent int) string {
+	var out strings.Builder
+	for _, raw := range raws {
+		var el RichTextElement
+		if err := json.Unmarshal(raw, &el); err != nil {
+			continue
+		}
+		switch el.Type {
+		case "rich_text_section":
+			out.WriteString(renderInline(el.Elements, resolver))
+			out.WriteString("\n")
+		case "rich_text_preformatted":
+			out.WriteString("```\n")
+			out.WriteString(renderInline(el.Elements, resolver))
+			out.WriteString("\n```\n")
+		case "rich_text_quote":
+			body := renderInline(el.Elements, resolver)
+			for _, line := range strings.Split(strings.TrimRight(body, "\n"), "\n") {
+				out.WriteString("> ")
+				out.WriteString(line)
+				out.WriteString("\n")
+			}
+		case "rich_text_list":
+			indent := strings.Repeat("  ", baseIndent+el.Indent)
+			start := el.Offset
+			for i, itemRaw := range el.Elements {
+				var item RichTextElement
+				if err := json.Unmarshal(itemRaw, &item); err != nil {
+					continue
+				}
+				// List items are rich_text_section elements
+				body := renderInline(item.Elements, resolver)
+				out.WriteString(indent)
+				if el.ListStyle == "ordered" {
+					fmt.Fprintf(&out, "%d. ", start+i+1)
+				} else {
+					out.WriteString("- ")
+				}
+				out.WriteString(body)
+				out.WriteString("\n")
+			}
+		default:
+			// Unknown sub-block type — drop silently.
+		}
+	}
+	return out.String()
+}
+
+// renderInline walks the inline elements of a rich_text_section.
+func renderInline(raws []json.RawMessage, resolver *UserResolver) string {
+	var out strings.Builder
+	for _, raw := range raws {
+		var el RichTextElement
+		if err := json.Unmarshal(raw, &el); err != nil {
+			continue
+		}
+		switch el.Type {
+		case "text":
+			out.WriteString(applyTextStyle(el.Text, el.TextStyle))
+		case "link":
+			if el.Text != "" {
+				fmt.Fprintf(&out, "[%s](%s)", el.Text, el.URL)
+			} else {
+				fmt.Fprintf(&out, "<%s>", el.URL)
+			}
+		case "user":
+			name := resolver.Resolve(el.UserID)
+			fmt.Fprintf(&out, "@%s", name)
+		case "channel":
+			fmt.Fprintf(&out, "<#%s>", el.ChannelID)
+		case "emoji":
+			fmt.Fprintf(&out, ":%s:", el.Name)
+		case "broadcast":
+			fmt.Fprintf(&out, "@%s", el.Range)
+		case "usergroup":
+			fmt.Fprintf(&out, "<!subteam^%s>", el.UsergroupID)
+		case "date":
+			if el.Fallback != "" {
+				out.WriteString(el.Fallback)
+			} else if el.Timestamp > 0 {
+				out.WriteString(time.Unix(el.Timestamp, 0).UTC().Format(time.RFC3339))
+			}
+			// Slack's date-format token language ({date_num}, {date}, etc.)
+			// is intentionally not interpreted — Format is read for JSON
+			// round-tripping but ignored during rendering.
+		default:
+			// Unknown inline element — drop silently so fall-back to
+			// m.Text can win if every element is unknown.
+		}
+	}
+	return out.String()
+}
+
+func applyTextStyle(s string, style *RichTextStyle) string {
+	if style == nil || s == "" {
+		return s
+	}
+	if style.Code {
+		s = "`" + s + "`"
+	}
+	if style.Bold {
+		s = "**" + s + "**"
+	}
+	if style.Italic {
+		s = "*" + s + "*"
+	}
+	if style.Strike {
+		s = "~~" + s + "~~"
+	}
+	return s
+}
--- a/internal/client/blocks.go
+++ b/internal/client/blocks.go
@@ -138,11 +138,21 @@ func renderRichText(raws []json.RawMessage, resolver *UserResolver, baseIndent i
 		}
 		switch el.Type {
 		case "rich_text_section":
-			out.WriteString(renderInline(el.Elements, resolver))
+			inline := renderInline(el.Elements, resolver)
+			if inline == "" {
+				// Skip empty sections (all inline elements unknown) so they
+				// don't emit blank indented lines in thread's multi-line view.
+				continue
+			}
+			out.WriteString(inline)
 			out.WriteString("\n")
 		case "rich_text_preformatted":
+			// Trim trailing newlines on the inner content so we emit
+			// "```\n<body>\n```" even if the inline text carries its own
+			// trailing "\n" (valid Slack payload).
+			inline := strings.TrimRight(renderInline(el.Elements, resolver), "\n")
 			out.WriteString("```\n")
-			out.WriteString(renderInline(el.Elements, resolver))
+			out.WriteString(inline)
 			out.WriteString("\n```\n")
 		case "rich_text_quote":
 			body := renderInline(el.Elements, resolver)
--- a/internal/client/blocks.go
+++ b/internal/client/blocks.go
@@ -9,12 +9,39 @@ import (
 
 // Block is a top-level Slack Block Kit block. Only rich_text is rendered
 // today; other types (section, image, context, etc.) are preserved in the
-// JSON representation but yield empty output from RenderBlocks, letting
-// callers fall back to the message's m.Text plain-text fallback.
+// JSON representation via raw-bytes round-tripping, and yield empty output
+// from RenderBlocks so callers fall back to the message's m.Text
+// plain-text fallback.
 type Block struct {
 	Type     string            `json:"type"`
 	BlockID  string            `json:"block_id,omitempty"`
 	Elements []json.RawMessage `json:"elements,omitempty"`
+
+	// raw holds the verbatim JSON bytes captured during unmarshal so that
+	// non-rich_text blocks (section.text, image.image_url, header.text,
+	// etc.) round-trip through --output json without losing fields the
+	// typed struct doesn't model.
+	raw json.RawMessage
+}
+
+// UnmarshalJSON captures the raw bytes alongside the typed fields so that
+// non-rich_text blocks can be re-emitted without loss.
+func (b *Block) UnmarshalJSON(data []byte) error {
+	b.raw = append(b.raw[:0], data...)
+	type alias Block
+	return json.Unmarshal(data, (*alias)(b))
+}
+
+// MarshalJSON emits the original raw bytes for non-rich_text blocks (which
+// carry type-specific fields the struct does not model) and the typed form
+// for rich_text blocks (which preserves polymorphic-style re-emission
+// through RichTextElement.MarshalJSON).
+func (b Block) MarshalJSON() ([]byte, error) {
+	if b.Type != "rich_text" && len(b.raw) > 0 {
+		return b.raw, nil
+	}
+	type alias Block
+	return json.Marshal(alias(b))
 }
 
 // RichTextElement is one element inside a rich_text block. The zero value
@@ -147,15 +174,23 @@ func renderRichText(raws []json.RawMessage, resolver *UserResolver, baseIndent i
 			out.WriteString(inline)
 			out.WriteString("\n")
 		case "rich_text_preformatted":
-			// Trim trailing newlines on the inner content so we emit
-			// "```\n<body>\n```" even if the inline text carries its own
-			// trailing "\n" (valid Slack payload).
-			inline := strings.TrimRight(renderInline(el.Elements, resolver), "\n")
+			// Preformatted blocks strip inline styles — markdown markers
+			// inside a fenced code block would render as literal text in
+			// downstream consumers.
+			inline := strings.TrimRight(renderInlineRaw(el.Elements, resolver), "\n")
+			if inline == "" {
+				continue
+			}
 			out.WriteString("```\n")
 			out.WriteString(inline)
 			out.WriteString("\n```\n")
 		case "rich_text_quote":
 			body := renderInline(el.Elements, resolver)
+			if body == "" {
+				// Skip quotes with no renderable content so we don't emit
+				// a phantom "> " line from the Split(\"\", \"\\n\") edge.
+				continue
+			}
 			for _, line := range strings.Split(strings.TrimRight(body, "\n"), "\n") {
 				out.WriteString("> ")
 				out.WriteString(line)
@@ -164,21 +199,28 @@ func renderRichText(raws []json.RawMessage, resolver *UserResolver, baseIndent i
 		case "rich_text_list":
 			indent := strings.Repeat("  ", baseIndent+el.Indent)
 			start := el.Offset
-			for i, itemRaw := range el.Elements {
+			// Track rendered items separately from the loop index so that
+			// skipped items (malformed JSON or empty body) don't leave
+			// numbering gaps in ordered lists.
+			rendered := 0
+			for _, itemRaw := range el.Elements {
 				var item RichTextElement
 				if err := json.Unmarshal(itemRaw, &item); err != nil {
 					continue
 				}
-				// List items are rich_text_section elements
 				body := renderInline(item.Elements, resolver)
+				if body == "" {
+					continue
+				}
 				out.WriteString(indent)
 				if el.ListStyle == "ordered" {
-					fmt.Fprintf(&out, "%d. ", start+i+1)
+					fmt.Fprintf(&out, "%d. ", start+rendered+1)
 				} else {
 					out.WriteString("- ")
 				}
 				out.WriteString(body)
 				out.WriteString("\n")
+				rendered++
 			}
 		default:
 			// Unknown sub-block type — drop silently.
@@ -187,8 +229,19 @@ func renderRichText(raws []json.RawMessage, resolver *UserResolver, baseIndent i
 	return out.String()
 }
 
+// renderInlineRaw is like renderInline but strips inline text styles, for
+// use inside rich_text_preformatted where markdown markers would render as
+// literal text rather than formatting.
+func renderInlineRaw(raws []json.RawMessage, resolver *UserResolver) string {
+	return renderInlineWithStyle(raws, resolver, false)
+}
+
 // renderInline walks the inline elements of a rich_text_section.
 func renderInline(raws []json.RawMessage, resolver *UserResolver) string {
+	return renderInlineWithStyle(raws, resolver, true)
+}
+
+func renderInlineWithStyle(raws []json.RawMessage, resolver *UserResolver, applyStyles bool) string {
 	var out strings.Builder
 	for _, raw := range raws {
 		var el RichTextElement
@@ -197,7 +250,11 @@ func renderInline(raws []json.RawMessage, resolver *UserResolver) string {
 		}
 		switch el.Type {
 		case "text":
-			out.WriteString(applyTextStyle(el.Text, el.TextStyle))
+			if applyStyles {
+				out.WriteString(applyTextStyle(el.Text, el.TextStyle))
+			} else {
+				out.WriteString(el.Text)
+			}
 		case "link":
 			if el.Text != "" {
 				fmt.Fprintf(&out, "[%s](%s)", el.Text, el.URL)
--- a/internal/client/blocks.go
+++ b/internal/client/blocks.go
@@ -33,9 +33,12 @@ func (b *Block) UnmarshalJSON(data []byte) error {
 }
 
 // MarshalJSON emits the original raw bytes for non-rich_text blocks (which
-// carry type-specific fields the struct does not model) and the typed form
-// for rich_text blocks (which preserves polymorphic-style re-emission
-// through RichTextElement.MarshalJSON).
+// carry type-specific fields the struct does not model). For rich_text
+// blocks it re-emits via the typed alias; Elements is []json.RawMessage
+// so each element's original bytes pass through unchanged, which is what
+// preserves the polymorphic "style" field on rich_text_list items and
+// inline text runs. RichTextElement.UnmarshalJSON/MarshalJSON are only
+// invoked when a caller explicitly unmarshals an individual element.
 func (b Block) MarshalJSON() ([]byte, error) {
 	if b.Type != "rich_text" && len(b.raw) > 0 {
 		return b.raw, nil
--- a/internal/cmd/messages/history.go
+++ b/internal/cmd/messages/history.go
@@ -64,7 +64,10 @@ func runHistory(channel string, opts *historyOptions, c *client.Client) error {
 	resolver := client.NewUserResolver(c)
 	for _, m := range messages {
 		ts := formatTimestamp(m.TS)
-		text := truncate(resolver.ResolveMentions(m.Text), 80)
+		// Compact view — truncation inherently flattens, so the
+		// blocks-vs-text distinction doesn't matter here.
+		body, _ := messageBody(m, resolver)
+		text := truncate(body, 80)
 		name := resolver.Resolve(m.User)
 		edited := ""
 		if m.Edited != nil {
--- a/internal/cmd/messages/history.go
+++ b/internal/cmd/messages/history.go
@@ -71,6 +71,9 @@ func runHistory(channel string, opts *historyOptions, c *client.Client) error {
 			edited = " [edited]"
 		}
 		output.Printf("[%s] %s: %s%s\n", ts, name, text, edited)
+		if files := renderFiles(m.Files); files != "" {
+			output.Printf("%s", files)
+		}
 	}
 
 	return nil
--- a/internal/cmd/messages/messages.go
+++ b/internal/cmd/messages/messages.go
@@ -120,7 +120,12 @@ func renderFiles(files []client.File) string {
 }
 
 // humanSize formats a byte count as "411 B", "12.3 KB", "4.5 MB", etc.
+// Slack occasionally returns size=-1 for certain snippet types; clamp
+// negatives to 0 rather than render "-1 B".
 func humanSize(n int64) string {
+	if n < 0 {
+		n = 0
+	}
 	const (
 		kb = 1024
 		mb = 1024 * kb
--- a/internal/cmd/messages/messages.go
+++ b/internal/cmd/messages/messages.go
@@ -8,6 +8,8 @@ import (
 	"unicode/utf8"
 
 	"github.com/spf13/cobra"
+
+	"github.com/open-cli-collective/slack-chat-api/internal/client"
 )
 
 // NewCmd creates the messages command with all subcommands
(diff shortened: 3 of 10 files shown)


rianjs:
**TL;DR:** Coverage is genuinely thorough for a feature of this complexity. The `RenderBlocks` walker, polymorphic unmarshaling, `renderFiles`, and `humanSize` all have focused unit tests, and nine integration-level tests drive the thread/history commands end-to-end. A handful of gaps could let silent regressions slip through, but none are show-stoppers.

## Coverage

### 🟥 Must-fix

**`messageBody` with a nil resolver passed to `ResolveMentions` (text path)**
`messageBody` calls `resolver.ResolveMentions(m.Text)` when blocks render empty. `ResolveMentions` now guards against a nil receiver, but there is no test that exercises `messageBody` itself with a nil resolver on the text-only branch. The resolver nil-guard in `resolver.go` is tested only via the blocks path (`TestRenderBlocks_NilResolverDoesNotPanic`). If the guard were accidentally removed from `ResolveMentions`, the text-only fallback in `runThread`/`runHistory` would panic in the field and no test would catch it.

**`MarshalJSON` round-trip not tested**
`RichTextElement.MarshalJSON` is new code that re-emits the polymorphic `style` field. No test unmarshals a `RichTextElement`, marshals it back, and verifies the output JSON. If `MarshalJSON` silently dropped `style` (e.g. returned `alias` without the `Style` field), `TestRunThread_JSONIncludesBlocks` would still pass because it only checks the block `Type`, not the inner element style.

### 🟨 Worth adding

**`indentContinuation` unit test**
`indentContinuation` is a pure function with three observable behaviors: interior `\n` → `\n\t`, trailing `\n` trimmed, and no-op on a single line. It is tested only indirectly via the large `TestRunThread_MultilineBlocksIndentContinuationLines` integration test. A focused unit test would be cheaper and would pin the trailing-newline trim contract directly.

**`rich_text_preformatted` with multiline content**
`TestRenderBlocks_Preformatted` uses a single-line body and checks only that `` ``` `` and the body text appear somewhere in the output. The preformatted renderer emits `\n` inside the fences; a multiline body (e.g. `"line1\nline2"`) could expose unexpected double-newlines. Worth one additional case.

**`rich_text_quote` with a single-line body (no interior newline)**
`TestRenderBlocks_Quote` only tests `"line one\nline two"`. A body with no `\n` follows a slightly different code path (`strings.Split` returns a one-element slice) that is not explicitly covered.

**`rich_text_list` item with malformed inner JSON**
In `renderRichText`, each list item is re-unmarshaled (`json.Unmarshal(itemRaw, &item)`) and silently skipped on error. No test exercises a list where one item is invalid JSON — the skip behavior is untested and could silently produce a truncated list.

**`broadcast` range values other than `here`**
`TestRenderBlocks_InlineElementTypes` uses only `"here"`. The implementation emits `@<range>` for any value; `channel` and `everyone` are untested but are common in real messages.

**`date` element with zero `Timestamp` and no `Fallback`**
The implementation silently emits nothing when both `Timestamp == 0` and `Fallback == ""`. This is a plausible real input (malformed Slack payload) and the silent-drop behavior is not verified.

**`TestRunHistory_TextIncludesFileHints` does not assert file hint renders in full when text is truncated**
The test asserts `assert.NotContains(t, out, "...F0ATD4WJ70D")` (file ID not in a truncated fragment), but does not verify the complete hint string `\t[file] IW Trailer.pdf (pdf, 59.2 KB) — slck files download F0ATD4WJ70D` is present verbatim. A subtle truncation bug that clips the download command after a certain point would not be caught.

### 🟩 Nice-to-have

**`humanSize` negative input** — the switch has no `n < 0` guard; a garbage API response sends it to the `< kb` case and prints a negative number. Low-probability but trivial to verify.

**`applyTextStyle` with an empty string** — `applyTextStyle` short-circuits on `s == ""`. A test with `style != nil, s == ""` would confirm the empty string passes through unstyled (currently only covered implicitly).

**Unknown top-level block alongside a known one** — `TestRenderBlocks_UnknownBlockTypeDropped` uses a slice with only one unknown block. A mixed slice (unknown + `rich_text`) is not tested; the known block should still render correctly.

## Strengths

- All four block-level container types (`rich_text_section`, `rich_text_list`, `rich_text_preformatted`, `rich_text_quote`) have dedicated tests.
- Every inline element type in the implementation (`text`, `link`, `user`, `channel`, `emoji`, `broadcast`, `usergroup`, `date`) is covered by at least one assertion.
- Style polymorphism (string vs. object vs. absent) is tested directly at the `UnmarshalJSON` level in addition to via `RenderBlocks`.
- The blocks-over-text priority and the fall-back-to-text-when-blocks-unrenderable paths are both explicitly covered at the integration level, including the edge case where unknown inline elements cause the whole section to render empty.
- `humanSize` boundary values (1023 B vs 1024 KB, exact GB) are all present.
- `renderFiles` title-vs-name precedence and tab-prefix contract are pinned with exact-string assertions.
- JSON output exposing `blocks` is verified end-to-end.

rianjs:
Reviewed this against the business goals from #141 and the architectural direction discussed for the fix. No blocking findings from this pass.

What aligns well:
- File hints now satisfy the core business need: text output explicitly surfaces attachments with a fully actionable `slck files download <id>` hint in both `thread` and `history`.
- Rich-text recovery follows the intended contract: blocks win when renderable, fallback text still wins when rendering would be empty, and `thread` vs `history` keep their different UX behaviors.
- The implementation split matches the architectural direction: `internal/client` owns Slack-shape parsing/rendering (`Block`, `RichTextElement`, `RenderBlocks`), while `internal/cmd/messages` owns command-specific presentation decisions (`messageBody`, flatten vs multiline, file-hint placement).
- Test coverage is strong for the intended scope, and I was able to reproduce the green checks locally with `make test` and `make lint`.

Residual caveat, but not a blocker for this PR:
- Unsupported top-level non-`rich_text` block-only messages will still render blank if Slack does not populate `m.Text`. That is the remaining blind spot created by the intentional “drop unknown blocks so fallback text can win” rule. For this PR’s scope, that tradeoff looks reasonable because the business problem here was specifically rich-text table-paste recovery, not generic Block Kit rendering.

rianjs:
Addressed in the latest commit (`test(messages): close coverage gaps flagged by TDD review`).

**Must-fix:**
- ✅ `TestMessageBody_NilResolverTextPath` — direct test of the text-only path with a nil resolver.
- ✅ `TestRichTextElement_MarshalJSONRoundTrip` — verifies the string variant, object variant, and absent-style case, all round-tripping through unmarshal → marshal → unmarshal.

**Worth-adding (all addressed):**
- ✅ `TestIndentContinuation` — direct unit test, 6 cases
- ✅ `TestRenderBlocks_PreformattedMultiline` — `line1\nline2\nline3` body
- ✅ `TestRenderBlocks_QuoteSingleLine` — no interior `\n` path
- ✅ `TestRenderBlocks_ListItemMalformedJSONIsSkipped` — asserts exactly one bullet line survives
- ✅ `TestRenderBlocks_BroadcastRanges` — `here`, `channel`, `everyone`
- ✅ `TestRenderBlocks_DateZeroTimestampNoFallback`
- ✅ `TestRunHistory_TextIncludesFileHints` — now pins the full hint line verbatim

**Nice-to-have picked up:**
- ✅ `TestRenderBlocks_UnknownBlockAlongsideKnown` — mixed slice renders the known block, drops the unknown.

Skipped the other two nice-to-haves (negative `humanSize`, empty-string `applyTextStyle`) — the negative-size path emits "-N B" which is harmless cosmetic behavior against a malformed payload, and `applyTextStyle` with `s == ""` is a one-line guard that the compiler enforces.

609 tests pass, lint clean. Leaving the PR as-is pending the architectural review.

internal/cmd/messages/history.go
```diff
@@ -64,13 +64,19 @@ func runHistory(channel string, opts *historyOptions, c *client.Client) error {
 	resolver := client.NewUserResolver(c)
 	for _, m := range messages {
 		ts := formatTimestamp(m.TS)
-		text := truncate(resolver.ResolveMentions(m.Text), 80)
+		// Compact view — truncation inherently flattens, so the
```
monit-reviewer, line 67: 🟡 **Medium** (harness-engineering:architecture-reviewer): Multi-line block bodies are not flattened before truncation. When messageBody returns rendered block content with multiple rich_text_section elements (e.g. "line1\nline2"), truncate(body, 80) preserves embedded newlines, which then land raw inside output.Printf("[%s] %s: %s%s\n", ...), breaking the compact single-line history format. The inline comment "truncation inherently flattens" is incorrect. Fix: use the fromBlocks return value and call flatten(body) before truncate when fromBlocks is true (mirroring thread.go's text path). TestRunHistory_BlocksTruncatedToEightyChars uses a single-section 200-char body and does not catch this.

internal/cmd/messages/messages_test.go
```diff
@@ -1691,3 +1691,573 @@ func TestRunSend_FileUpload_LongTextAllowed(t *testing.T) {
...
+	out := captureTextOutput(t, func() {
+		require.NoError(t, runThread("C123", "1234567890.123456", opts, c))
+	})
+	assert.Contains(t, out, "line1\n\tline2", "expected continuation line indented with tab")
+}
+
+func TestRunThread_PreservesFlattenForTextOnlyMultilineMessage(t *testing.T) {
+	// Regression guard: plain-text messages with newlines should still be
+	// flattened when no blocks are present.
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
monit-reviewer, line 2194: 🔵 **Low** (harness-engineering:architecture-reviewer): Missing test for runHistory with a multi-section rich_text block. Existing TestRunHistory_BlocksTruncatedToEightyChars uses a single-section body with no interior newlines, so it cannot catch the compact-view newline regression in history.go. Add a sub-test with two rich_text_section elements (each under 80 chars) asserting the output remains on one line.

internal/client/blocks.go
```diff
@@ -0,0 +1,242 @@
...
+			continue
+		}
+		switch el.Type {
+		case "rich_text_section":
+			out.WriteString(renderInline(el.Elements, resolver))
+			out.WriteString("\n")
+		case "rich_text_preformatted":
+			out.WriteString("```\n")
+			out.WriteString(renderInline(el.Elements, resolver))
+			out.WriteString("\n```\n")
```
monit-reviewer, line 189: 🔵 **Low** (harness-engineering:enforcement-reviewer): renderRichText unconditionally appends '\n' after a rich_text_section even when renderInline returns empty (forward-compat scenario where all inline elements are unknown types). TrimRight in RenderBlocks strips trailing newlines, but a leading or interior empty section produces a blank line — in thread view, indentContinuation then emits '\n\t' right after the [ts] user: header, creating a visible artifact. Skip the trailing newline when renderInline returns empty.

internal/client/blocks.go
```diff
@@ -0,0 +1,242 @@
...
+		return ""
+	}
+	var out strings.Builder
+	for _, b := range blocks {
+		if b.Type != "rich_text" {
+			// Non-rich_text blocks (section, image, context, etc.) are
+			// dropped so the caller falls back to m.Text, which Slack
+			// populates as a plain-text equivalent for these cases.
+			continue
+		}
```
monit-reviewer, line 155: 🔵 **Low** (harness-engineering:architecture-reviewer): rich_text_preformatted rendering writes "\n```\n" unconditionally after renderInline. If an inline text element inside the preformatted block carries a trailing \n (valid Slack payload), the closing fence is preceded by a blank line: "```\ncode\n\n```". Add strings.TrimRight(..., "\n") around the renderInline result before appending the closing fence.

internal/cmd/messages/messages.go
```diff
@@ -61,6 +63,71 @@ func truncate(s string, maxLen int) string {
...
+		b.WriteString(name)
+		b.WriteString(" (")
+		b.WriteString(f.Filetype)
+		b.WriteString(", ")
+		b.WriteString(humanSize(f.Size))
+		b.WriteString(") — slck files download ")
+		b.WriteString(f.ID)
+		b.WriteString("\n")
+	}
+	return b.String()
```
monit-reviewer, line 119: 🔵 **Low** (harness-engineering:enforcement-reviewer): renderFiles does not guard against an empty f.Filetype. When Slack omits the filetype field, the output becomes '\t[file] name (, 411 B) — slck files download F123' with a cosmetically broken parenthetical. Either omit the type clause when empty or substitute a fallback like 'unknown'.

monit-reviewer (dismissed):
## Automated PR Review



**Reviewed commit:** `01009d3`

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:architecture-reviewer | 3 |
| harness-engineering:enforcement-reviewer | 2 |

<details>
<summary><strong>harness-engineering:architecture-reviewer (3 findings)</strong></summary>

**⚠️ Should Fix** - `internal/cmd/messages/history.go:67`
> Multi-line block bodies are not flattened before truncation. When messageBody returns rendered block content with multiple rich_text_section elements (e.g. "line1\nline2"), truncate(body, 80) preserves embedded newlines, which then land raw inside output.Printf("[%s] %s: %s%s\n", ...), breaking the compact single-line history format. The inline comment "truncation inherently flattens" is incorrect. Fix: use the fromBlocks return value and call flatten(body) before truncate when fromBlocks is true (mirroring thread.go's text path). TestRunHistory_BlocksTruncatedToEightyChars uses a single-section 200-char body and does not catch this.

**💡 Suggestion** - `internal/cmd/messages/messages_test.go:2094`
> Missing test for runHistory with a multi-section rich_text block. Existing TestRunHistory_BlocksTruncatedToEightyChars uses a single-section body with no interior newlines, so it cannot catch the compact-view newline regression in history.go. Add a sub-test with two rich_text_section elements (each under 80 chars) asserting the output remains on one line.

**💡 Suggestion** - `internal/client/blocks.go:125`
> rich_text_preformatted rendering writes "\n```\n" unconditionally after renderInline. If an inline text element inside the preformatted block carries a trailing \n (valid Slack payload), the closing fence is preceded by a blank line: "```\ncode\n\n```". Add strings.TrimRight(..., "\n") around the renderInline result before appending the closing fence.

</details>

<details>
<summary><strong>harness-engineering:enforcement-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/client/blocks.go:146`
> renderRichText unconditionally appends '\n' after a rich_text_section even when renderInline returns empty (forward-compat scenario where all inline elements are unknown types). TrimRight in RenderBlocks strips trailing newlines, but a leading or interior empty section produces a blank line — in thread view, indentContinuation then emits '\n\t' right after the [ts] user: header, creating a visible artifact. Skip the trailing newline when renderInline returns empty.

**💡 Suggestion** - `internal/cmd/messages/messages.go:109`
> renderFiles does not guard against an empty f.Filetype. When Slack omits the filetype field, the output becomes '\t[file] name (, 411 B) — slck files download F123' with a cosmetically broken parenthetical. Either omit the type clause when empty or substitute a fallback like 'unknown'.

</details>

*4 info-level observations excluded. Run with `--verbose` to include.*

*3 PR discussion threads considered.*

---
<details>
<summary>Completed in 5m 24s | $2.59</summary>

| Field | Value |
|---|---|
| Reviewers | hybrid-synthesis, security:code-auditor, harness-engineering:knowledge-reviewer, harness-engineering:enforcement-reviewer, harness-engineering:architecture-reviewer |
| Engine | claude |
| Duration | 5m 24s wall · 10m 17s compute (Reviewers: 4m 28s · Synthesis: 45s) |
| Cost | $2.59 |
| Tokens | 413.8k in / 30.6k out |
| Turns | 11 |

</details>

internal/cmd/messages/history.go
rianjs: Disagree — the existing `truncate()` in `internal/cmd/messages/messages.go:57-63` already does `strings.ReplaceAll(s, "\n", " ")` as its first statement, unconditionally, before any length check. So multi-line block bodies ARE flattened here regardless of length. Added `TestRunHistory_MultiSectionBlocksStayOnOneLine` as a regression guard to lock that in.

internal/cmd/messages/messages_test.go
rianjs: Added `TestRunHistory_MultiSectionBlocksStayOnOneLine` with two rich_text_section elements; asserts the output is a single line.

internal/client/blocks.go
rianjs: Fixed — empty `rich_text_section` elements are now skipped (`continue` when `renderInline` returns `"""). Added `TestRenderBlocks_EmptySectionDoesNotEmitBlankLine`.

internal/client/blocks.go
rianjs: Fixed — `strings.TrimRight(..., "\n")` on the inline body before the closing fence. Added `TestRenderBlocks_PreformattedInnerTrailingNewlineTrimmed`.

internal/cmd/messages/messages.go
rianjs: Fixed — the type clause is omitted entirely when `f.Filetype` is empty (no `(, size)` cosmetic artifact). Added an "empty filetype drops the type clause" sub-test to `TestRenderFiles`.

internal/cmd/messages/history.go
monit-reviewer: **Settled**: monit-reviewer claimed multi-line block bodies aren't flattened before truncation. rianjs pointed out that truncate() already does ReplaceAll newlines unconditionally and added a regression test.

internal/cmd/messages/messages_test.go
monit-reviewer: **Settled**: monit-reviewer requested a test for multi-section rich_text blocks in runHistory. rianjs added TestRunHistory_MultiSectionBlocksStayOnOneLine.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer flagged that empty rich_text_section elements produce blank-line artifacts. rianjs fixed by skipping empty sections and added a regression test.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer flagged a blank line before the closing fence in rich_text_preformatted rendering. rianjs added strings.TrimRight and a regression test.

internal/cmd/messages/messages.go
monit-reviewer: **Settled**: monit-reviewer flagged a cosmetic artifact when f.Filetype is empty in renderFiles. rianjs fixed by omitting the type clause and added a sub-test.

internal/client/blocks.go
```diff
@@ -0,0 +1,252 @@
...
+	"fmt"
+	"strings"
+	"time"
+)
+
+// Block is a top-level Slack Block Kit block. Only rich_text is rendered
+// today; other types (section, image, context, etc.) are preserved in the
+// JSON representation but yield empty output from RenderBlocks, letting
+// callers fall back to the message's m.Text plain-text fallback.
+type Block struct {
```
monit-reviewer, line 15: 🟡 **Medium** (harness-engineering:knowledge-reviewer): Block struct captures only type/block_id/elements. Slack block types like section, header, image, input, file, video carry content in other fields (text, fields, image_url, alt_text, etc.) that are not modeled. Messages with these block types lose their content fields on JSON round-trip, contradicting the PR description's claim that --output json exposes blocks 'unchanged'. Section blocks are the most common case in practice. Consider storing Message.Blocks as []json.RawMessage and decoding only the rich_text shape during rendering.

internal/client/blocks.go
```diff
@@ -0,0 +1,252 @@
...
+		if err := json.Unmarshal(raw, &el); err != nil {
+			continue
+		}
+		switch el.Type {
+		case "rich_text_section":
+			inline := renderInline(el.Elements, resolver)
+			if inline == "" {
+				// Skip empty sections (all inline elements unknown) so they
+				// don't emit blank indented lines in thread's multi-line view.
+				continue
```
monit-reviewer, line 175: 🔵 **Low** (harness-engineering:architecture-reviewer): rich_text_quote and rich_text_preformatted have no empty-body guard analogous to rich_text_section. If renderInline returns "" (all inline elements unknown), rich_text_quote emits a phantom "> " line (strings.Split("", "\n") yields [""], loop runs once) and rich_text_preformatted emits "```\n\n```". Apply the same skip-on-empty guard used in the section branch.

internal/client/blocks.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+			out.WriteString(inline)
+			out.WriteString("\n")
+		case "rich_text_preformatted":
+			// Trim trailing newlines on the inner content so we emit
+			// "```\n<body>\n```" even if the inline text carries its own
+			// trailing "\n" (valid Slack payload).
+			inline := strings.TrimRight(renderInline(el.Elements, resolver), "\n")
```
monit-reviewer: 🔵 **Low** (harness-engineering:architecture-reviewer): rich_text_list renders phantom empty bullets/numbers for list items with no renderable inline content. The loop calls renderInline(item.Elements, resolver) without a guard, so an item with all-unknown elements produces '- \n' instead of being skipped, paralleling the section-case empty guard.

internal/client/blocks.go
```diff
@@ -0,0 +1,252 @@
...
+					continue
+				}
+				// List items are rich_text_section elements
+				body := renderInline(item.Elements, resolver)
+				out.WriteString(indent)
+				if el.ListStyle == "ordered" {
+					fmt.Fprintf(&out, "%d. ", start+i+1)
+				} else {
+					out.WriteString("- ")
+				}
```
monit-reviewer, line 223: 🔵 **Low** (harness-engineering:enforcement-reviewer): Ordered-list numbering gaps when an item fails to unmarshal. The loop index still increments on continue, so a bad item at index 1 causes subsequent items to be numbered 3, 4, … instead of 2, 3, …. The existing malformed-item test only covers bullet lists where the gap is invisible; an ordered-list case would expose the off-by-one.

internal/client/blocks.go
```diff
@@ -0,0 +1,252 @@
...
+		switch el.Type {
+		case "rich_text_section":
+			inline := renderInline(el.Elements, resolver)
+			if inline == "" {
+				// Skip empty sections (all inline elements unknown) so they
+				// don't emit blank indented lines in thread's multi-line view.
+				continue
+			}
+			out.WriteString(inline)
+			out.WriteString("\n")
```
monit-reviewer, line 178: 🔵 **Low** (harness-engineering:knowledge-reviewer): Inside the rich_text_preformatted case, renderInline is called without stripping style markers. If an inline text element within a preformatted block carries bold/italic/code/strike style flags (structurally valid per Slack Block Kit), applyTextStyle would emit markdown syntax inside the triple-backtick fence, producing malformed output. Slack's current UI does not generate this combination but the renderer has no guard.

internal/cmd/messages/messages.go
```diff
@@ -61,6 +63,75 @@ func truncate(s string, maxLen int) string {
...
+// Any trailing newline is trimmed first to avoid a dangling tab.
+func indentContinuation(s string) string {
+	s = strings.TrimRight(s, "\n")
+	return strings.ReplaceAll(s, "\n", "\n\t")
+}
+
+// renderFiles returns one tab-indented "[file] ..." line per attachment, each
+// terminated with "\n". Returns "" when files is empty. The format gives a
+// reader (human or agent) enough context to invoke `slck files download <id>`.
+func renderFiles(files []client.File) string {
```
monit-reviewer, line 89: 🔵 **Low** (harness-engineering:knowledge-reviewer): renderFiles does not handle the case where both f.Title and f.Name are empty (possible for Slack snippets before a name is set, or certain file-sharing events). Output becomes '\t[file]  (<type>, <size>) — slck files download <id>' with a blank display name and a double space. No test covers the empty-name/empty-title combination.

internal/client/blocks.go
```diff
@@ -0,0 +1,252 @@
+package client
```
monit-reviewer, line 1: 🔵 **Low** (harness-engineering:architecture-reviewer): RenderBlocks/renderRichText/renderInline/applyTextStyle are presentation functions placed in the client package, which is otherwise a Slack API wrapper. This couples rendering to the transport layer; a second output format would force duplication or cross-package coupling. Rendering logic would fit more naturally in internal/cmd/messages or a dedicated internal/output subpackage.

monit-reviewer (dismissed):
## Automated PR Review



**Reviewed commit:** `974b00e` | **Previous:** `01009d3` (incremental)

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:architecture-reviewer | 3 |
| harness-engineering:enforcement-reviewer | 1 |
| harness-engineering:knowledge-reviewer | 3 |

<details>
<summary><strong>harness-engineering:architecture-reviewer (3 findings)</strong></summary>

**💡 Suggestion** - `internal/client/blocks.go:145`
> rich_text_quote and rich_text_preformatted have no empty-body guard analogous to rich_text_section. If renderInline returns "" (all inline elements unknown), rich_text_quote emits a phantom "> " line (strings.Split("", "\n") yields [""], loop runs once) and rich_text_preformatted emits "```\n\n```". Apply the same skip-on-empty guard used in the section branch.

**💡 Suggestion** - `internal/client/blocks.go:153`
> rich_text_list renders phantom empty bullets/numbers for list items with no renderable inline content. The loop calls renderInline(item.Elements, resolver) without a guard, so an item with all-unknown elements produces '- \n' instead of being skipped, paralleling the section-case empty guard.

**💡 Suggestion** - `internal/client/blocks.go:1`
> RenderBlocks/renderRichText/renderInline/applyTextStyle are presentation functions placed in the client package, which is otherwise a Slack API wrapper. This couples rendering to the transport layer; a second output format would force duplication or cross-package coupling. Rendering logic would fit more naturally in internal/cmd/messages or a dedicated internal/output subpackage.

</details>

<details>
<summary><strong>harness-engineering:enforcement-reviewer (1 findings)</strong></summary>

**💡 Suggestion** - `internal/client/blocks.go:179`
> Ordered-list numbering gaps when an item fails to unmarshal. The loop index still increments on continue, so a bad item at index 1 causes subsequent items to be numbered 3, 4, … instead of 2, 3, …. The existing malformed-item test only covers bullet lists where the gap is invisible; an ordered-list case would expose the off-by-one.

</details>

<details>
<summary><strong>harness-engineering:knowledge-reviewer (3 findings)</strong></summary>

**⚠️ Should Fix** - `internal/client/blocks.go:14`
> Block struct captures only type/block_id/elements. Slack block types like section, header, image, input, file, video carry content in other fields (text, fields, image_url, alt_text, etc.) that are not modeled. Messages with these block types lose their content fields on JSON round-trip, contradicting the PR description's claim that --output json exposes blocks 'unchanged'. Section blocks are the most common case in practice. Consider storing Message.Blocks as []json.RawMessage and decoding only the rich_text shape during rendering.

**💡 Suggestion** - `internal/client/blocks.go:148`
> Inside the rich_text_preformatted case, renderInline is called without stripping style markers. If an inline text element within a preformatted block carries bold/italic/code/strike style flags (structurally valid per Slack Block Kit), applyTextStyle would emit markdown syntax inside the triple-backtick fence, producing malformed output. Slack's current UI does not generate this combination but the renderer has no guard.

**💡 Suggestion** - `internal/cmd/messages/messages.go:89`
> renderFiles does not handle the case where both f.Title and f.Name are empty (possible for Slack snippets before a name is set, or certain file-sharing events). Output becomes '\t[file]  (<type>, <size>) — slck files download <id>' with a blank display name and a double space. No test covers the empty-name/empty-title combination.

</details>

*8 PR discussion threads considered.*

---
<details>
<summary>Completed in 6m 16s | $3.27</summary>

| Field | Value |
|---|---|
| Mode | Re-review · Cycle 2 · Session resumed |
| Reviewers | hybrid-synthesis, security:code-auditor, harness-engineering:knowledge-reviewer, harness-engineering:enforcement-reviewer, harness-engineering:architecture-reviewer |
| Engine | claude |
| Duration | 6m 16s wall · 13m 17s compute (Reviewers: 4m 50s · Synthesis: 47s) |
| Cost | $3.27 |
| Tokens | 607.6k in / 27.7k out |
| Turns | 16 |

</details>

internal/client/blocks.go
rianjs: Fixed — `Block` now captures raw bytes in `UnmarshalJSON` and emits them verbatim via `MarshalJSON` for non-rich_text blocks, preserving `section.text`, `image.image_url`, `header.text`, etc. rich_text blocks still go through typed marshal so `RichTextElement.MarshalJSON` preserves polymorphic style. Added `TestBlock_JSONFidelityForNonRichTextBlocks` and `TestBlock_JSONFidelityForRichTextUsesTypedMarshal`.

internal/client/blocks.go
rianjs: Fixed — both `rich_text_quote` and `rich_text_preformatted` now skip when the inline body renders empty. Added `TestRenderBlocks_QuoteWithEmptyBodySkipped` and `TestRenderBlocks_PreformattedWithEmptyBodySkipped`.

internal/client/blocks.go
rianjs: Fixed — list items with empty rendered body are skipped. Added `TestRenderBlocks_ListItemWithEmptyBodySkipped` asserting exactly one bullet line when one item's inline content is all-unknown.

internal/client/blocks.go
rianjs: Fixed — track rendered items with a separate counter so skipped items don't leave numbering gaps. Added `TestRenderBlocks_OrderedListSkipPreservesNumbering` with a malformed middle item; remaining items number 1, 2 (not 1, 3).

internal/client/blocks.go
rianjs: Fixed — added `renderInlineRaw` (a no-style variant of `renderInline`) used inside `rich_text_preformatted`. Inline text with `TextStyle` flags now renders as literal characters inside the fence rather than leaking markdown markers. Added `TestRenderBlocks_PreformattedStripsInlineStyles`.

internal/cmd/messages/messages.go
rianjs: Fixed — `renderFiles` now falls back to `f.ID` when both `Title` and `Name` are empty, so the line never renders as `[file]  (...)` with a blank display name. Added "empty name and title falls back to file ID" sub-test.

internal/client/blocks.go
rianjs: Disagree — the first-pass architectural review (issue-comment-4276152296) explicitly approved this split: "`internal/client` owns Slack-shape parsing/rendering (`Block`, `RichTextElement`, `RenderBlocks`), while `internal/cmd/messages` owns command-specific presentation decisions (`messageBody`, flatten vs multiline, file-hint placement)." Moving `RenderBlocks` out of `internal/client` would either split the rich-text types from their renderer across a package boundary (coupled by `json.RawMessage`) or invert the dependency direction. The current split treats `RenderBlocks` as a Slack-shape decoder — the last mile of parsing — rather than presentation; `messageBody`/`flatten`/`truncate` in `internal/cmd/messages` are the presentation layer. If a second output format emerges, consumers can walk `Message.Blocks` directly; `RenderBlocks` is one convenience walker among potentially many.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer flagged that the Block struct only captured rich_text fields, losing content on JSON round-trip for section/header/image/etc. rianjs fixed it by capturing raw bytes in UnmarshalJSON and emitting verbatim via MarshalJSON for non-rich_text blocks, with added fidelity tests.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer noted rich_text_quote and rich_text_preformatted lacked the empty-body guard, producing phantom '> ' or empty fenced blocks. rianjs added skip-on-empty guards and tests for both.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer flagged rich_text_list rendering phantom empty bullets when items had all-unknown inline content. rianjs added a guard to skip empty items and a test.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer flagged ordered-list numbering gaps when items failed to unmarshal. rianjs fixed it with a separate rendered-item counter and added a test verifying contiguous numbering.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer flagged that inline style markers weren't stripped inside rich_text_preformatted, risking markdown leaking into code fences. rianjs added renderInlineRaw for no-style rendering in preformatted blocks plus a test.

internal/cmd/messages/messages.go
monit-reviewer: **Settled**: monit-reviewer flagged renderFiles producing a blank display name when both Title and Name are empty. rianjs added a fallback to file ID and a sub-test.

internal/client/blocks.go
monit-reviewer: **Settled**: monit-reviewer suggested moving rendering functions out of internal/client to decouple from the transport layer. rianjs disagreed, citing an earlier architectural review that approved the split treating RenderBlocks as Slack-shape decoding rather than presentation.

internal/client/blocks.go
```diff
@@ -0,0 +1,309 @@
...
+// UnmarshalJSON captures the raw bytes alongside the typed fields so that
+// non-rich_text blocks can be re-emitted without loss.
+func (b *Block) UnmarshalJSON(data []byte) error {
+	b.raw = append(b.raw[:0], data...)
+	type alias Block
+	return json.Unmarshal(data, (*alias)(b))
+}
+
+// MarshalJSON emits the original raw bytes for non-rich_text blocks (which
+// carry type-specific fields the struct does not model) and the typed form
```
monit-reviewer: 🔵 **Low** (harness-engineering:enforcement-reviewer): Block.MarshalJSON's comment claims rich_text blocks round-trip 'through RichTextElement.MarshalJSON', but Block.Elements is []json.RawMessage — marshaling alias(b) emits those raw bytes directly and RichTextElement.MarshalJSON is never invoked on this path. The fidelity test passes only because the raw captured bytes already contain the style field. Either update the comment to describe the actual mechanism (raw-byte passthrough of elements) or switch Elements to typed RichTextElement so the stated contract holds.

internal/client/blocks.go
```diff
@@ -0,0 +1,309 @@
...
+		return b.raw, nil
+	}
+	type alias Block
+	return json.Marshal(alias(b))
+}
+
+// RichTextElement is one element inside a rich_text block. The zero value
+// plus the non-nil Elements signal a container (rich_text_section,
+// rich_text_list, rich_text_preformatted, rich_text_quote). Otherwise the
+// element is an inline primitive (text, link, user, channel, emoji,
```
monit-reviewer, line 53: 🔵 **Low** (harness-engineering:knowledge-reviewer): Block.MarshalJSON uses the typed struct for rich_text blocks and only models type/block_id/elements. Unknown top-level fields Slack may send on a rich_text block (future additions like `border`, `fallback`, etc.) are silently dropped on --output json round-trip, unlike non-rich_text blocks which preserve everything via raw bytes. Consider capturing raw bytes for rich_text blocks too (and decoding only during rendering), or document the limitation.

internal/client/blocks.go
```diff
@@ -0,0 +1,309 @@
...
+				out.WriteString(body)
+				out.WriteString("\n")
+				rendered++
+			}
+		default:
+			// Unknown sub-block type — drop silently.
+		}
+	}
+	return out.String()
+}
```
monit-reviewer, line 233: 🔵 **Low** (harness-engineering:enforcement-reviewer): renderInlineWithStyle's 'text' case emits el.Text without calling ResolveMentions. If Slack emits a literal <@UXXXXX> inside a plain text inline element (observed in some legacy/edited payloads), it renders raw, while the m.Text fallback in messageBody resolves it. Same mention renders differently depending on the path. Consider resolving mentions inside text-type inline elements as well, or document that only dedicated `user`/`channel`/`usergroup` inline types get resolution.

internal/cmd/messages/messages.go
```diff
@@ -61,6 +63,81 @@ func truncate(s string, maxLen int) string {
...
+			// snippets or certain file-sharing events. Fall back to the
+			// file ID so the line never renders as "[file]  (...)".
+			name = f.ID
+		}
+		b.WriteString("\t[file] ")
+		b.WriteString(name)
+		b.WriteString(" (")
+		// Slack occasionally omits filetype; drop the type clause rather
+		// than emit "(, 411 B)".
+		if f.Filetype != "" {
```
monit-reviewer, line 110: 🔵 **Low** (harness-engineering:knowledge-reviewer): humanSize does not guard against negative input. If Slack returns size=-1 (observed with certain snippet types), output is '-1 B'. A `if n < 0 { n = 0 }` guard at the top would prevent confusing output at no semantic cost.

internal/cmd/messages/thread.go
```diff
@@ -66,13 +66,25 @@ func runThread(channel, threadTS string, opts *threadOptions, c *client.Client)
...
+		} else {
+			// Plain-text fallback keeps existing single-line behavior.
+			text = flatten(body)
+		}
 		name := resolver.Resolve(m.User)
 		edited := ""
 		if m.Edited != nil {
 			edited = " [edited]"
 		}
 		output.Printf("[%s] %s: %s%s\n", ts, name, text, edited)
```
monit-reviewer, line 95: 🔵 **Low** (harness-engineering:enforcement-reviewer): When fromBlocks=true and the rendered body has interior newlines (e.g., two-section block produces 'line1\n\tline2'), the [edited] marker appends to the final continuation line: '[ts] user: line1\n\tline2 [edited]'. Visually the tag appears to annotate only the last line rather than the whole message. Consider appending [edited] to the header line or the end of the first line instead.

monit-reviewer (approved):
## Automated PR Review



**Reviewed commit:** `4092342` | **Previous:** `974b00e` (incremental)

> **Approved** with 5 non-blocking suggestions below. Address at your discretion.

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:enforcement-reviewer | 3 |
| harness-engineering:knowledge-reviewer | 2 |

<details>
<summary><strong>harness-engineering:enforcement-reviewer (3 findings)</strong></summary>

**💡 Suggestion** - `internal/client/blocks.go:36`
> Block.MarshalJSON's comment claims rich_text blocks round-trip 'through RichTextElement.MarshalJSON', but Block.Elements is []json.RawMessage — marshaling alias(b) emits those raw bytes directly and RichTextElement.MarshalJSON is never invoked on this path. The fidelity test passes only because the raw captured bytes already contain the style field. Either update the comment to describe the actual mechanism (raw-byte passthrough of elements) or switch Elements to typed RichTextElement so the stated contract holds.

**💡 Suggestion** - `internal/client/blocks.go:230`
> renderInlineWithStyle's 'text' case emits el.Text without calling ResolveMentions. If Slack emits a literal <@UXXXXX> inside a plain text inline element (observed in some legacy/edited payloads), it renders raw, while the m.Text fallback in messageBody resolves it. Same mention renders differently depending on the path. Consider resolving mentions inside text-type inline elements as well, or document that only dedicated `user`/`channel`/`usergroup` inline types get resolution.

**💡 Suggestion** - `internal/cmd/messages/thread.go:84`
> When fromBlocks=true and the rendered body has interior newlines (e.g., two-section block produces 'line1\n\tline2'), the [edited] marker appends to the final continuation line: '[ts] user: line1\n\tline2 [edited]'. Visually the tag appears to annotate only the last line rather than the whole message. Consider appending [edited] to the header line or the end of the first line instead.

</details>

<details>
<summary><strong>harness-engineering:knowledge-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/client/blocks.go:50`
> Block.MarshalJSON uses the typed struct for rich_text blocks and only models type/block_id/elements. Unknown top-level fields Slack may send on a rich_text block (future additions like `border`, `fallback`, etc.) are silently dropped on --output json round-trip, unlike non-rich_text blocks which preserve everything via raw bytes. Consider capturing raw bytes for rich_text blocks too (and decoding only during rendering), or document the limitation.

**💡 Suggestion** - `internal/cmd/messages/messages.go:110`
> humanSize does not guard against negative input. If Slack returns size=-1 (observed with certain snippet types), output is '-1 B'. A `if n < 0 { n = 0 }` guard at the top would prevent confusing output at no semantic cost.

</details>

*15 PR discussion threads considered.*

---
<details>
<summary>Completed in 5m 06s | $2.59</summary>

| Field | Value |
|---|---|
| Mode | Re-review · Cycle 3 · Session resumed |
| Reviewers | hybrid-synthesis, security:code-auditor, harness-engineering:knowledge-reviewer, harness-engineering:enforcement-reviewer, harness-engineering:architecture-reviewer |
| Engine | claude |
| Duration | 5m 06s wall · 10m 24s compute (Reviewers: 3m 35s · Synthesis: 42s) |
| Cost | $2.59 |
| Tokens | 397.2k in / 22.7k out |
| Turns | 10 |

</details>

Issue #147, opened 2026-04-29
feat(files): expose file refs in search and add files get


rianjs:
## Problem

`slck search files` text output is not command-ready. It shows file labels, type, user, date, and title, but omits the Slack file ID that `slck files download` needs.

Agents should not need to fall back to `--output json` to continue from search to get/download.

Current likely workflow:

```sh
slck search files "quarterly report"
slck search files "quarterly report" -o json | jq ... # needed to find file id
slck files download F0ABC123
```

Desired workflow:

```sh
slck search files "quarterly report"
slck files get F0ABC123
slck files download F0ABC123
```

## Product Decision

Expose Slack file IDs as command-ready refs in text search output. For files, the canonical ref is just the Slack file ID:

```text
F0ABC123
```

Default JSON output remains upstream-shaped and must not add rendered/opinionated fields. This fix belongs in text output and command surface.

## Scope

- Change `search files` text output to include a `REF` column containing the file ID.
- Change the file section of `search all` text output to include the same `REF` column.
- Prefer pipe-delimited, unpadded search output for agent/token efficiency:

```text
REF | NAME | TYPE | USER | CREATED | TITLE
F0ABC123 | report.pdf | pdf | U123 | 2026-04-28 | Quarterly report
```

- Keep free-text/title fields at the end where practical and flatten internal newlines.
- Make pipe handling deterministic in text cells, e.g. replace literal `|` in cell values with a safe substitute or otherwise document stable parsing.
- Add `slck files get <file-id-or-url>` for plaintext file metadata inspection.
- Keep `slck files download <file-id-or-url>` as the command that writes bytes to disk.
- Reuse the existing file ID / URL parsing from `files download` where possible.

## Suggested `files get` Text Output

Keep it compact and command-forward:

```text
ID: F0ABC123
Name: report.pdf
Title: Quarterly report
Type: pdf
Size: 59.2 KB
Download: slck files download F0ABC123
```

Include permalink/download URL only if current command conventions already expose links in text output; otherwise prefer the next `slck` command over raw URLs.

## Non-Goals

- Do not download file bytes from `files get`.
- Do not require JSON parsing to discover the file ID.
- Do not mutate JSON output or add derived JSON fields.
- Do not solve file-to-message/thread breadcrumbs in this first pass unless `files.info` already exposes the needed share metadata cleanly.

## Follow-Up Breadcrumb Idea

A later issue can cover zooming out from a file to the message/thread where it was shared:

```text
[...]

PR #148, opened 2026-04-29
feat(messages): add message refs and read command for search-to-read workflow

### LINKED ISSUE #146: feat(messages): add message refs and read command for search-to-read workflow
## Problem

`slck search messages` now renders useful message text, but the text output still does not give agents a command-ready way to read the full message/thread without falling back to `--output json` or doing a second history lookup.

The common workflow should be two commands:

```sh
slck search messages "product-aware backend" --in "#engineering-team" --include-bots
slck --as-user messages read <message-ref>
```

Today, search output shows a human timestamp and truncated text, but not the composite Slack locator needed for `conversations.replies`: `channel_id + ts`.

## Product Decision

Expose a compact, command-ready message ref in text output. Treat a Slack message ref as:

```text
<channel_id>/<message_ts>
```

Example:

```text
C02DF3BEUGN/1777469221.721439
```

This is the API-level composite identity for a message in a channel. The CLI should hide the fact that lower-level APIs need two separate fields.

Default JSON output remains upstream-shaped and must not add rendered/opinionated fields. This fix belongs in text output and command parsing.

## Scope

- Add a small message ref parser/formatter, e.g. `MessageRef{ChannelID, TS}`.
- Accept at least `<channel_id>/<ts>`.
- Consider accepting Slack permalinks too, since timestamp normalization already supports permalink timestamps.
- Add `slck messages read <message-ref>` as the plaintext read command.
- `messages read` should render the natural readable unit:
  - parent message + replies when the ref points to a thread parent;
  - parent thread when the ref points to a reply, if Slack/API data makes that practical;
  - otherwise the single message response from `conversations.replies` is acceptable for the first version.
- Keep using the shared message renderer from #143/#144.
- Change `search messages` text output to include a `REF` column.
- Change the message section of `search all` text output to include the same `REF` column.
- Prefer pipe-delimited, unpadded search output for agent/token efficiency:

```text
REF | CHANNEL | USER | WHEN | TEXT
C02DF3BEUGN/1777469221.721439 | engineering | mon-bot | 2026-04-29 10:17 | Product-aware backend...
```

- Keep `TEXT` as the final column and flatten internal newlines.
- Make pipe handling deterministic in text cells, e.g. replace literal `|` in cell values with a safe substitute or otherwise document stable parsing.
- Add a short footer hint after non-empty search results:

```text
Read: slck --as-user messages read <REF>
```

## Non-Goals

- Do not mutate JSON `text` or add renderer-derived JSON fields.
- Do not require users/agents to parse `--output json` for this workflow.
- Do not preserve old aligned-table text formatting if pipe output better serves the workflow.
- Do not add multiple new read/context commands unless one command with flags cannot cover the use case.

## Open Design Notes

Potential follow-up flags for `messages read` once the base workflow lands:

```sh
slck messages read <ref> --single
slck messages read <ref> --around 3
```

These cover zooming into one message or zooming out to nearby channel context without creating separate low-value command surface area.

## Acceptance Criteria

- A search result row contains a command-ready `REF` without using JSON.
- The displayed ref can be pasted directly into `slck --as-user messages read <REF>`.
- `messages read` returns plaintext using the shared renderer, including block/attachment/file text surfaces.
- `search all` message rows expose the same ref as `search messages`.
- Tests cover block-only message search results, message ref parsing/formatting, and the read command path.

## Related

- #143
- #144
- #145

rianjs:
## Summary
- New `internal/messageref` package: parses `<channel_id>/<ts>` and Slack permalinks into a `Ref` and formats canonically. Reuses `validate` for ts/channel validation; URL grammar lives in `messageref`.
- `slck messages read <ref>`: reads the conversation thread for a ref via `conversations.replies`. Wraps `not_in_channel` with a `--as-user` hint; other errors pass through untouched.
- `search messages` and `search all` (messages section) now use a pipe-delimited table with a leading REF column (channel-id/ts) and a footer hint pointing at `slck --as-user messages read <REF>`. Hint suppressed when no message rows are present (file-only `search all` results don't see it).
- New `output.SearchTable(headers, rows, lastColMaxRunes)` helper — single formatting path for search rows across both #146 and #147 (files). Owns newline flattening, pipe-escape (`|` -> `¦`), and rune-based last-column truncation. Mismatched row widths pad/trim deterministically.
- Thread rendering loop extracted to a shared `renderMessageList` so `messages thread` and `messages read` cannot drift on edited markers, file lines, or user resolution.

JSON output is unchanged: no rendered text fields, no synthetic `ref` field — text output is the only product surface for the new ref + footer hint.

Closes #146

## Test plan
- [x] `go test ./...` — 699 passed
- [x] `make lint` — clean
- [x] `messageref` parse/format covers API form, permalink (with query string), p-prefixed ts, invalid channel/ts/missing slash
- [x] `output.SearchTable` covers pipe-escape, newline flatten, rune truncation, width mismatch
- [x] `search messages` REF column uses channel ID (not name); footer present with results, absent when empty
- [x] `search all` footer suppressed for file-only results
- [x] `messages read` covers success, permalink input, `not_in_channel` hint, unrelated error pass-through, JSON output, invalid ref

### DIFF
--- a/internal/cmd/messages/read.go
+++ b/internal/cmd/messages/read.go
@@ -1,0 +1,81 @@
+package messages
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/open-cli-collective/slack-chat-api/internal/client"
+	"github.com/open-cli-collective/slack-chat-api/internal/messageref"
+	"github.com/open-cli-collective/slack-chat-api/internal/output"
+)
+
+type readOptions struct {
+	limit int
+}
+
+func newReadCmd() *cobra.Command {
+	opts := &readOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "read <message-ref>",
+		Short: "Read a message thread by ref",
+		Long: `Read the conversation thread for a message ref.
+
+A message ref is the composite identity for a message in a channel:
+  <channel_id>/<ts>           e.g. C02DF3BEUGN/1777469221.721439
+  Slack permalink             e.g. https://workspace.slack.com/archives/C02DF3BEUGN/p1777469221721439
+
+Refs are emitted as the REF column by 'slck search messages' and 'slck search all'.
+
+Returns whatever Slack's conversations.replies returns for the ref. When the ref
+points at a thread parent, that's the parent plus all replies. Some channels
+require user-token semantics; pair with --as-user when search returned the ref.
+
+Examples:
+  slck messages read C02DF3BEUGN/1777469221.721439
+  slck --as-user messages read C02DF3BEUGN/1777469221.721439`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runRead(args[0], opts, nil)
+		},
+	}
+
+	cmd.Flags().IntVar(&opts.limit, "limit", 100, "Maximum messages to return")
+
+	return cmd
+}
+
+func runRead(input string, opts *readOptions, c *client.Client) error {
+	ref, err := messageref.Parse(input)
+	if err != nil {
+		return err
+	}
+
+	if c == nil {
+		c, err = client.New()
+		if err != nil {
+			return err
+		}
+	}
+
+	messages, err := c.GetThreadReplies(ref.ChannelID, ref.TS, opts.limit, "")
+	if err != nil {
+		if client.IsSlackError(err, "not_in_channel") {
+			return fmt.Errorf("%w\nHint: try `slck --as-user messages read %s` — search-derived refs typically require user-token access", err, ref)
+		}
+		return err
+	}
+
+	if output.IsJSON() {
+		return output.PrintJSON(messages)
+	}
+
+	if len(messages) == 0 {
+		output.Println("No messages found for ref")
+		return nil
+	}
+
+	renderMessageList(messages, client.NewUserResolver(c))
+	return nil
+}
--- a/internal/cmd/messages/read.go
+++ b/internal/cmd/messages/read.go
@@ -62,7 +62,7 @@ func runRead(input string, opts *readOptions, c *client.Client) error {
 	messages, err := c.GetThreadReplies(ref.ChannelID, ref.TS, opts.limit, "")
 	if err != nil {
 		if client.IsSlackError(err, "not_in_channel") {
-			return fmt.Errorf("%w\nHint: try `slck --as-user messages read %s` — search-derived refs typically require user-token access", err, ref)
+			return fmt.Errorf("%w; try `slck --as-user messages read %s` — search-derived refs typically require user-token access", err, ref)
 		}
 		return err
 	}
--- a/internal/cmd/search/all.go
+++ b/internal/cmd/search/all.go
@@ -141,6 +141,7 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 		paging := result.Messages.Paging
 		output.Printf("\nPage %d of %d (showing %d of %d messages)\n",
 			paging.Page, paging.Pages, len(result.Messages.Matches), paging.Total)
+		output.Println("Read: slck --as-user messages read <REF>")
 	}
 
 	// Add spacing between sections
@@ -165,10 +166,6 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 		paging := result.Files.Paging
 		output.Printf("\nPage %d of %d (showing %d of %d files)\n",
 			paging.Page, paging.Pages, len(result.Files.Matches), paging.Total)
-	}
-
-	if hasMessages {
-		output.Println("\nRead: slck --as-user messages read <REF>")
 	}
 
 	return nil
--- a/internal/cmd/search/all.go
+++ b/internal/cmd/search/all.go
@@ -4,6 +4,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/open-cli-collective/slack-chat-api/internal/client"
+	"github.com/open-cli-collective/slack-chat-api/internal/messageref"
 	"github.com/open-cli-collective/slack-chat-api/internal/output"
 )
 
@@ -132,7 +133,7 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 				Files:       m.Files,
 			}, nil).Body
 			when := formatTimestamp(m.TS)
-			ref := m.Channel.ID + "/" + m.TS
+			ref := messageref.Ref{ChannelID: m.Channel.ID, TS: m.TS}.String()
 			rows = append(rows, []string{ref, m.Channel.Name, m.Username, when, body})
 		}
 		output.SearchTable(headers, rows, 60)
--- a/internal/cmd/search/all.go
+++ b/internal/cmd/search/all.go
@@ -122,7 +122,7 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 	if hasMessages {
 		output.Printf("=== Messages (%d total) ===\n\n", result.Messages.Total)
 
-		headers := []string{"CHANNEL", "USER", "TIMESTAMP", "TEXT"}
+		headers := []string{"REF", "CHANNEL", "USER", "WHEN", "TEXT"}
 		rows := make([][]string, 0, len(result.Messages.Matches))
 		for _, m := range result.Messages.Matches {
 			body := client.RenderMessage(client.MessageContent{
@@ -131,11 +131,11 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 				Attachments: m.Attachments,
 				Files:       m.Files,
 			}, nil).Body
-			text := truncateText(body, 60)
-			ts := formatTimestamp(m.TS)
-			rows = append(rows, []string{m.Channel.Name, m.Username, ts, text})
+			when := formatTimestamp(m.TS)
+			ref := m.Channel.ID + "/" + m.TS
+			rows = append(rows, []string{ref, m.Channel.Name, m.Username, when, body})
 		}
-		output.Table(headers, rows)
+		output.SearchTable(headers, rows, 60)
 
 		paging := result.Messages.Paging
 		output.Printf("\nPage %d of %d (showing %d of %d messages)\n",
@@ -164,6 +164,10 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 		paging := result.Files.Paging
 		output.Printf("\nPage %d of %d (showing %d of %d files)\n",
 			paging.Page, paging.Pages, len(result.Files.Matches), paging.Total)
+	}
+
+	if hasMessages {
+		output.Println("\nRead: slck --as-user messages read <REF>")
 	}
 
 	return nil
--- a/internal/cmd/search/messages.go
+++ b/internal/cmd/search/messages.go
@@ -126,7 +126,7 @@ func runSearchMessages(query string, opts *messagesOptions, c *client.Client) er
 
 	output.Printf("Found %d messages matching \"%s\"\n\n", result.Messages.Total, query)
 
-	headers := []string{"CHANNEL", "USER", "TIMESTAMP", "TEXT"}
+	headers := []string{"REF", "CHANNEL", "USER", "WHEN", "TEXT"}
 	rows := make([][]string, 0, len(result.Messages.Matches))
 	for _, m := range result.Messages.Matches {
 		body := client.RenderMessage(client.MessageContent{
@@ -135,15 +135,16 @@ func runSearchMessages(query string, opts *messagesOptions, c *client.Client) er
 			Attachments: m.Attachments,
 			Files:       m.Files,
 		}, nil).Body
-		text := truncateText(body, 60)
-		ts := formatTimestamp(m.TS)
-		rows = append(rows, []string{m.Channel.Name, m.Username, ts, text})
+		when := formatTimestamp(m.TS)
+		ref := m.Channel.ID + "/" + m.TS
+		rows = append(rows, []string{ref, m.Channel.Name, m.Username, when, body})
 	}
-	output.Table(headers, rows)
+	output.SearchTable(headers, rows, 60)
 
 	paging := result.Messages.Paging
 	output.Printf("\nPage %d of %d (showing %d of %d results)\n",
 		paging.Page, paging.Pages, len(result.Messages.Matches), paging.Total)
+	output.Println("Read: slck --as-user messages read <REF>")
 
 	return nil
 }
--- a/internal/cmd/search/messages.go
+++ b/internal/cmd/search/messages.go
@@ -9,6 +9,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/open-cli-collective/slack-chat-api/internal/client"
+	"github.com/open-cli-collective/slack-chat-api/internal/messageref"
 	"github.com/open-cli-collective/slack-chat-api/internal/output"
 )
 
@@ -136,7 +137,7 @@ func runSearchMessages(query string, opts *messagesOptions, c *client.Client) er
 			Files:       m.Files,
 		}, nil).Body
 		when := formatTimestamp(m.TS)
-		ref := m.Channel.ID + "/" + m.TS
+		ref := messageref.Ref{ChannelID: m.Channel.ID, TS: m.TS}.String()
 		rows = append(rows, []string{ref, m.Channel.Name, m.Username, when, body})
 	}
 	output.SearchTable(headers, rows, 60)
--- a/internal/cmd/search/search_test.go
+++ b/internal/cmd/search/search_test.go
@@ -943,3 +943,101 @@ func TestSearchMessages_WithSortOptions(t *testing.T) {
 		t.Errorf("unexpected error: %v", err)
 	}
 }
+
+// TestRunSearchMessages_RefColumnAndFooter verifies the REF column uses
+// channel ID (not channel name) and that the footer hint is emitted when
+// matches are present.
+func TestRunSearchMessages_RefColumnAndFooter(t *testing.T) {
+	response := map[string]interface{}{
+		"ok": true,
+		"messages": map[string]interface{}{
+			"total":  1,
+			"paging": map[string]interface{}{"count": 20, "total": 1, "page": 1, "pages": 1},
+			"matches": []map[string]interface{}{{
+				"type":     "message",
+				"channel":  map[string]interface{}{"id": "C02DF3BEUGN", "name": "engineering"},
+				"user":     "U1",
+				"username": "alice",
+				"text":     "hello",
+				"ts":       "1777469221.721439",
+			}},
+		},
+	}
+	c, server := newTestClient(func(w http.ResponseWriter, _ *http.Request) {
+		_ = json.NewEncoder(w).Encode(response)
+	})
+	defer server.Close()
+
+	out := captureOutput(t, func() {
+		if err := runSearchMessages("hello", &messagesOptions{count: 20, page: 1, sort: "score", sortDir: "desc"}, c); err != nil {
+			t.Fatalf("runSearchMessages: %v", err)
+		}
+	})
+
+	if !strings.Contains(out, "REF | CHANNEL | USER | WHEN | TEXT") {
+		t.Errorf("missing REF header line; got:\n%s", out)
+	}
+	if !strings.Contains(out, "C02DF3BEUGN/1777469221.721439 | engineering |") {
+		t.Errorf("REF should be channel-id/ts; got:\n%s", out)
+	}
+	if !strings.Contains(out, "Read: slck --as-user messages read <REF>") {
+		t.Errorf("missing footer hint; got:\n%s", out)
+	}
+}
+
+func TestRunSearchMessages_NoFooterWhenEmpty(t *testing.T) {
+	response := map[string]interface{}{
+		"ok": true,
+		"messages": map[string]interface{}{
+			"total":   0,
+			"paging":  map[string]interface{}{"count": 20, "total": 0, "page": 1, "pages": 0},
+			"matches": []map[string]interface{}{},
+		},
+	}
+	c, server := newTestClient(func(w http.ResponseWriter, _ *http.Request) {
+		_ = json.NewEncoder(w).Encode(response)
+	})
+	defer server.Close()
+
+	out := captureOutput(t, func() {
+		if err := runSearchMessages("nope", &messagesOptions{count: 20, page: 1, sort: "score", sortDir: "desc"}, c); err != nil {
+			t.Fatalf("runSearchMessages: %v", err)
+		}
+	})
+
+	if strings.Contains(out, "Read: slck") {
+		t.Errorf("footer should not appear with no results; got:\n%s", out)
+	}
+}
+
+func TestRunSearchAll_FooterOnlyWithMessages(t *testing.T) {
+	// File-only result: footer hint must NOT appear.
+	response := map[string]interface{}{
+		"ok": true,
+		"files": map[string]interface{}{
+			"total":  1,
+			"paging": map[string]interface{}{"count": 20, "total": 1, "page": 1, "pages": 1},
+			"matches": []map[string]interface{}{{
+				"id":       "F1",
+				"name":     "report.pdf",
+				"title":    "report",
+				"filetype": "pdf",
+				"user":     "U1",
+				"created":  int64(1704067200),
+			}},
+		},
+	}
+	c, server := newTestClient(func(w http.ResponseWriter, _ *http.Request) {
+		_ = json.NewEncoder(w).Encode(response)
+	})
+	defer server.Close()
+
+	out := captureOutput(t, func() {
+		if err := runSearchAll("query", &allOptions{count: 20, page: 1, sort: "score", sortDir: "desc"}, c); err != nil {
+			t.Fatalf("runSearchAll: %v", err)
+		}
+	})
+	if strings.Contains(out, "Read: slck") {
+		t.Errorf("footer should be suppressed for file-only results; got:\n%s", out)
+	}
+}
--- a/internal/cmd/search/search_test.go
+++ b/internal/cmd/search/search_test.go
@@ -1041,3 +1041,33 @@ func TestRunSearchAll_FooterOnlyWithMessages(t *testing.T) {
 		t.Errorf("footer should be suppressed for file-only results; got:\n%s", out)
 	}
 }
+
+func TestRunSearchAll_FooterShownWithMessages(t *testing.T) {
+	response := map[string]interface{}{
+		"ok": true,
+		"messages": map[string]interface{}{
+			"total":  1,
+			"paging": map[string]interface{}{"count": 20, "total": 1, "page": 1, "pages": 1},
+			"matches": []map[string]interface{}{{
+				"type":     "message",
+				"channel":  map[string]interface{}{"id": "C123", "name": "general"},
+				"user":     "U1",
+				"username": "alice",
+				"text":     "hi",
+				"ts":       "1.0",
+			}},
+		},
+	}
+	c, server := newTestClient(func(w http.ResponseWriter, _ *http.Request) {
+		_ = json.NewEncoder(w).Encode(response)
+	})
+	defer server.Close()
+	out := captureOutput(t, func() {
+		if err := runSearchAll("hi", &allOptions{count: 20, page: 1, sort: "score", sortDir: "desc"}, c); err != nil {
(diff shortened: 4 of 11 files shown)


rianjs:
**Blocker**
None.

**Major**
- `internal/messageref/messageref.go:50` / `:67`: `Parse` delegates to `validate.ChannelID`, which only accepts `C`/`G`. Search/message refs should accept Slack conversation IDs that `messages thread` already treats as valid, including `D...` IM IDs. Otherwise a `REF` emitted from search for a DM is not command-ready. Use a message-ref/conversation-ID validator that accepts `C`/`G`/`D`, and add pair + permalink tests.

**Minor**
- `internal/cmd/search/messages.go:139` and `internal/cmd/search/all.go:135`: refs are formatted by string concatenation instead of the new `messageref.Ref.String()`. Keep the canonical representation owned by `messageref` so formatter/parser cannot drift.
- `internal/output/output.go:153`: docs say CR/newlines are collapsed to spaces, but CR is removed. Either replace `\r` with space too, or tighten the doc. Low impact, but this helper is intended to become shared infrastructure for #147.

**Nit**
None.

rianjs:
**Blocker**
None.

**Major**
None.

**Minor**
None.

**Nit**
None.

The fixes close the prior review points: message refs now cover `C/G/D` conversation IDs, formatting ownership is back in `messageref`, and `SearchTable` docs match behavior. PR #148 is aligned with the ticket and the two-ticket architecture.

rianjs:
## TDD Coverage Assessment

Test coverage on this PR is **adequate**. The added tests target real behavior (not implementation detail) across all four surfaces:

**`internal/messageref`** — table-driven `TestParse` covers C/G/D channel IDs, permalinks (incl. `?thread_ts=` query strings), `p`-prefixed ts normalization, whitespace trimming, plus the meaningful error cases (empty, missing slash, invalid channel prefix, user ID rejected, malformed permalink). Round-trip test pins canonical form.

**`internal/output.SearchTable`** — pipe-delimited format, newline flattening, `|` → `¦` substitution, last-column-only truncation, row/header length mismatch handling, empty headers no-panic, and a dedicated rune-vs-byte truncation test for multi-byte content.

**`messages read`** — happy path, permalink input, `not_in_channel` produces the `--as-user` hint with the original ref echoed, other errors do *not* get the hint, invalid ref short-circuits before the API, JSON output path.

**`search messages` / `search all`** — REF column uses channel ID (not name), footer hint appears with matches, suppressed when empty, and correctly suppressed for file-only `search all` results.

### Minor gaps (LOW priority, optional)

- No assertion that `messages read` honors the `--limit` flag value when calling `conversations.replies` (the success test hardcodes 100). A one-line query-param check would lock this in.
- `runRead` passes empty `cursor` to `GetThreadReplies`; pagination behavior isn't exercised but is also not new surface area introduced by this PR.

Neither gap is worth blocking on. Ship it.

rianjs:
Thanks — addressing the LOW gap: TestRunRead_Success now asserts `limit=42` propagates through to the `conversations.replies` query string. Skipping the pagination gap since GetThreadReplies pagination isn't new surface area in this PR.

internal/cmd/messages/read.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if c == nil {
+		c, err = client.New()
+		if err != nil {
+			return err
+		}
+	}
+
+	messages, err := c.GetThreadReplies(ref.ChannelID, ref.TS, opts.limit, "")
```
monit-reviewer, line 62: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): fmt.Errorf("%w\nHint: ...", err, ref) embeds a newline in the returned error string. Newlines in error values are non-idiomatic in Go and break structured logging, error chain comparison, and any caller that treats the error as a single-line string. The hint should be printed separately (e.g., output.Println) and a plain error returned, or the newline replaced with '; '.

internal/cmd/messages/read.go
```diff
@@ -0,0 +1,81 @@
...
+			return fmt.Errorf("%w\nHint: try `slck --as-user messages read %s` — search-derived refs typically require user-token access", err, ref)
+		}
+		return err
+	}
+
+	if output.IsJSON() {
+		return output.PrintJSON(messages)
+	}
+
+	if len(messages) == 0 {
```
monit-reviewer, line 74: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): The 'len(messages) == 0' branch (prints 'No messages found for ref' and returns nil) has no test coverage. conversations.replies can legitimately return an empty list for a deleted message or expired ref; the success+empty path should be exercised to guard against accidental removal or behavior change.

internal/cmd/search/all.go
```diff
@@ -166,5 +167,9 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 			paging.Page, paging.Pages, len(result.Files.Matches), paging.Total)
```
monit-reviewer, line 168: 🟡 **Medium** (harness-engineering:harness-enforcement-reviewer): When both messages and files are returned by 'search all', the footer hint 'Read: slck --as-user messages read <REF>' appears after the entire files section rather than immediately after the messages table. A user reading top-to-bottom sees the message table, then the files table, then the hint — the spatial disconnect makes it non-obvious which section the hint belongs to. Consider printing the hint directly after the messages paging line, before the files section begins.

internal/output/output.go
```diff
@@ -109,6 +109,68 @@ func Table(headers []string, rows [][]string) {
...
+
+	lastIdx := len(headers) - 1
+	for _, row := range rows {
+		cells := make([]string, len(headers))
+		for i := range headers {
+			var raw string
+			if i < len(row) {
+				raw = row[i]
+			}
+			c := sanitizeSearchCell(raw)
```
monit-reviewer, line 143: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): truncateRunes has a discontinuity at the boundary where max-3 == 1: max=3 returns 3 full runes (no ellipsis) while max=4 returns 1 rune + '...'. The jump from 3 visible characters to 1 visible character when the cap grows by one is surprising. The guard 'if max <= 3' was intended to avoid a zero- or negative-length prefix, but the threshold is off by one — max=4 still produces a very short visible prefix. Consider raising the guard to 'max < 6' or at minimum documenting the intended minimum usable cap.

internal/cmd/search/messages.go
```diff
@@ -135,15 +136,16 @@ func runSearchMessages(query string, opts *messagesOptions, c *client.Client) er
...
+		when := formatTimestamp(m.TS)
+		ref := messageref.Ref{ChannelID: m.Channel.ID, TS: m.TS}.String()
+		rows = append(rows, []string{ref, m.Channel.Name, m.Username, when, body})
 	}
-	output.Table(headers, rows)
+	output.SearchTable(headers, rows, 60)
 
 	paging := result.Messages.Paging
 	output.Printf("\nPage %d of %d (showing %d of %d results)\n",
 		paging.Page, paging.Pages, len(result.Messages.Matches), paging.Total)
```
monit-reviewer, line 147: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): Inconsistent footer spacing vs all.go: messages.go uses output.Println("Read: slck ...") (no leading blank line) while all.go uses output.Println("\nRead: slck ...") (with a leading blank line). This produces different vertical spacing between the paging line and the footer in the two commands, which is a visible inconsistency in CLI output style.

internal/cmd/search/search_test.go
```diff
@@ -943,3 +943,101 @@ func TestSearchMessages_WithSortOptions(t *testing.T) {
...
+		_ = json.NewEncoder(w).Encode(response)
+	})
+	defer server.Close()
+
+	out := captureOutput(t, func() {
+		if err := runSearchAll("query", &allOptions{count: 20, page: 1, sort: "score", sortDir: "desc"}, c); err != nil {
+			t.Fatalf("runSearchAll: %v", err)
+		}
+	})
+	if strings.Contains(out, "Read: slck") {
```
monit-reviewer, line 1040: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): TestRunSearchAll_FooterOnlyWithMessages only covers the negative case (file-only result, footer absent). There is no companion test verifying the footer IS present when 'search all' returns message matches. The all.go footer is guarded by 'if hasMessages', but this branch has no test coverage. A regression that accidentally dropped the hasMessages guard would go undetected.

internal/messageref/messageref.go
```diff
@@ -0,0 +1,89 @@
...
+		return Ref{}, fmt.Errorf("empty message ref")
+	}
+
+	if strings.HasPrefix(input, "http://") || strings.HasPrefix(input, "https://") {
+		m := permalinkPattern.FindStringSubmatch(input)
+		if m == nil {
+			return Ref{}, fmt.Errorf("invalid Slack permalink %q: expected https://<workspace>/archives/<channel>/p<digits>", input)
+		}
+		channel := m[1]
+		digits := m[2]
```
monit-reviewer, line 55: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): In the permalink branch, the channel is captured by [A-Z0-9]+ (no minimum-length constraint beyond 1 char) before being validated by conversationIDPattern (^[CGD][A-Z0-9]+$, minimum 2 chars). A malformed permalink such as '.../archives/C/p...' produces a validation error whose message reads "invalid conversation ID \"C\": must start with C, G, or D" — which is technically incorrect (C is a valid prefix letter) and will be confusing. The root cause is that conversationIDPattern implicitly enforces a minimum length of 2 but its error message describes only the prefix constraint.

internal/messageref/messageref.go
```diff
@@ -0,0 +1,89 @@
...
+
+// permalinkPattern matches Slack message permalinks.
+// Example: https://workspace.slack.com/archives/C02DF3BEUGN/p1777469221721439
+var permalinkPattern = regexp.MustCompile(`^https?://[^/]+/archives/([A-Z0-9]+)/p(\d{16})(?:\?.*)?$`)
+
+// Parse accepts either a permalink or "<channel_id>/<ts>" and returns a Ref.
+// The ts portion may be in API form (1234567890.123456) or p-prefixed form
+// (p1234567890123456); both are normalized to API form on the resulting Ref.
+func Parse(input string) (Ref, error) {
+	input = strings.TrimSpace(input)
```
monit-reviewer, line 44: 🔵 **Low** (harness-engineering:harness-self-documenting-code-reviewer): The permalink regex \d{16} hardcodes exactly 16 decimal digits for the p-prefixed timestamp. Slack's permalink format encodes a 10-digit epoch + 6-digit microseconds today, but the regex will silently reject valid permalinks if Slack ever emits a different number of digits. A looser match followed by structural validation (split at position 10 + validate.Timestamp) would be more resilient and consistent with how the <channel>/<ts> path is handled.

internal/output/output.go
```diff
@@ -109,6 +109,68 @@ func Table(headers []string, rows [][]string) {
 	}
 }
 
+// SearchTable writes pipe-delimited, unpadded rows for agent-friendly
+// search output: "h1 | h2 | h3" / "v1 | v2 | v3".
+//
+// Cell rules (callers pass raw strings only):
+//   - internal newlines are collapsed to single spaces; carriage returns are stripped
```
monit-reviewer, line 116: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): SearchTable always emits the header line regardless of whether rows is empty. Every call site that can reach SearchTable with 0 rows will produce a lonely header with no data rows. This is inconsistent with how output.Table behaves for the same case and may look wrong in command output.

monit-reviewer (dismissed):
## Automated PR Review



**Reviewed commit:** `beca3be`

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:harness-architecture-reviewer | 2 |
| harness-engineering:harness-enforcement-reviewer | 5 |
| harness-engineering:harness-knowledge-reviewer | 1 |
| harness-engineering:harness-self-documenting-code-reviewer | 1 |

<details>
<summary><strong>harness-engineering:harness-architecture-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/cmd/messages/read.go:62`
> fmt.Errorf("%w\nHint: ...", err, ref) embeds a newline in the returned error string. Newlines in error values are non-idiomatic in Go and break structured logging, error chain comparison, and any caller that treats the error as a single-line string. The hint should be printed separately (e.g., output.Println) and a plain error returned, or the newline replaced with '; '.

**💡 Suggestion** - `internal/cmd/messages/read.go:74`
> The 'len(messages) == 0' branch (prints 'No messages found for ref' and returns nil) has no test coverage. conversations.replies can legitimately return an empty list for a deleted message or expired ref; the success+empty path should be exercised to guard against accidental removal or behavior change.

</details>

<details>
<summary><strong>harness-engineering:harness-enforcement-reviewer (5 findings)</strong></summary>

**⚠️ Should Fix** - `internal/cmd/search/all.go:167`
> When both messages and files are returned by 'search all', the footer hint 'Read: slck --as-user messages read <REF>' appears after the entire files section rather than immediately after the messages table. A user reading top-to-bottom sees the message table, then the files table, then the hint — the spatial disconnect makes it non-obvious which section the hint belongs to. Consider printing the hint directly after the messages paging line, before the files section begins.

**💡 Suggestion** - `internal/output/output.go:143`
> truncateRunes has a discontinuity at the boundary where max-3 == 1: max=3 returns 3 full runes (no ellipsis) while max=4 returns 1 rune + '...'. The jump from 3 visible characters to 1 visible character when the cap grows by one is surprising. The guard 'if max <= 3' was intended to avoid a zero- or negative-length prefix, but the threshold is off by one — max=4 still produces a very short visible prefix. Consider raising the guard to 'max < 6' or at minimum documenting the intended minimum usable cap.

**💡 Suggestion** - `internal/cmd/search/messages.go:147`
> Inconsistent footer spacing vs all.go: messages.go uses output.Println("Read: slck ...") (no leading blank line) while all.go uses output.Println("\nRead: slck ...") (with a leading blank line). This produces different vertical spacing between the paging line and the footer in the two commands, which is a visible inconsistency in CLI output style.

**💡 Suggestion** - `internal/cmd/search/search_test.go:1040`
> TestRunSearchAll_FooterOnlyWithMessages only covers the negative case (file-only result, footer absent). There is no companion test verifying the footer IS present when 'search all' returns message matches. The all.go footer is guarded by 'if hasMessages', but this branch has no test coverage. A regression that accidentally dropped the hasMessages guard would go undetected.

**💡 Suggestion** - `internal/messageref/messageref.go:55`
> In the permalink branch, the channel is captured by [A-Z0-9]+ (no minimum-length constraint beyond 1 char) before being validated by conversationIDPattern (^[CGD][A-Z0-9]+$, minimum 2 chars). A malformed permalink such as '.../archives/C/p...' produces a validation error whose message reads "invalid conversation ID \"C\": must start with C, G, or D" — which is technically incorrect (C is a valid prefix letter) and will be confusing. The root cause is that conversationIDPattern implicitly enforces a minimum length of 2 but its error message describes only the prefix constraint.

</details>

<details>
<summary><strong>harness-engineering:harness-knowledge-reviewer (1 findings)</strong></summary>

**💡 Suggestion** - `internal/output/output.go:116`
> SearchTable always emits the header line regardless of whether rows is empty. Every call site that can reach SearchTable with 0 rows will produce a lonely header with no data rows. This is inconsistent with how output.Table behaves for the same case and may look wrong in command output.

</details>

<details>
<summary><strong>harness-engineering:harness-self-documenting-code-reviewer (1 findings)</strong></summary>

**💡 Suggestion** - `internal/messageref/messageref.go:44`
> The permalink regex \d{16} hardcodes exactly 16 decimal digits for the p-prefixed timestamp. Slack's permalink format encodes a 10-digit epoch + 6-digit microseconds today, but the regex will silently reject valid permalinks if Slack ever emits a different number of digits. A looser match followed by structural validation (split at position 10 + validate.Timestamp) would be more resilient and consistent with how the <channel>/<ts> path is handled.

</details>

*2 info-level observations excluded. Run with `--verbose` to include.*

*3 PR discussion threads considered.*

---
<details>
<summary>Completed in 3m 11s | $1.14</summary>

| Field | Value |
|---|---|
| Reviewers | hybrid-synthesis, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude |
| Duration | 3m 11s wall · 8m 56s compute (Reviewers: 2m 18s · Synthesis: 50s) |
| Cost | $1.14 |
| Tokens | 339.8k in / 34.6k out |
| Turns | 20 |

</details>

internal/cmd/messages/read.go
rianjs: Fixed in beca3be → next push: replaced `\n` with `; ` separator so the error stays single-line.

internal/cmd/messages/read.go
rianjs: Added TestRunRead_EmptyMessages covering the success+empty branch.

internal/cmd/search/all.go
rianjs: Moved the footer hint to immediately follow the messages paging line, before the files section.

internal/cmd/search/messages.go
rianjs: Both call sites now print the footer with no leading blank line, directly after the paging line — consistent.

internal/cmd/search/search_test.go
rianjs: Added TestRunSearchAll_FooterShownWithMessages as the positive companion.

internal/messageref/messageref.go
rianjs: Tightened permalink regex to `[A-Z0-9]{2,}` so a single-letter capture never reaches the validator.

internal/output/output.go
rianjs: Skipping. The discontinuity only affects caps in [4,5] which no caller passes (search uses 60). Adding a `min < 6` floor would change behavior for currently-impossible inputs; not worth widening the API contract for.

internal/messageref/messageref.go
rianjs: Skipping. Slack's permalink format has been stable at 10+6 digits for years, and the validate.Timestamp call after splitting at index 10 already enforces the digit shape. Loosening to a variable-length match would accept malformed permalinks the current regex correctly rejects.

internal/output/output.go
rianjs: Skipping. All current call sites gate `SearchTable` on `len(matches) > 0`, so the lonely-header case isn't reachable. Adding row-empty short-circuit logic would be defensive code for an unreachable path.

internal/cmd/messages/read.go
monit-reviewer: **Settled**: monit-reviewer flagged that embedding a newline in the error string via fmt.Errorf is non-idiomatic Go and breaks structured logging. rianjs fixed it by replacing the \n with a '; ' separator to keep the error single-line.

internal/cmd/messages/read.go
monit-reviewer: **Settled**: monit-reviewer noted the empty-messages branch lacked test coverage, leaving the success+empty path unguarded. rianjs added TestRunRead_EmptyMessages to cover it.

internal/cmd/search/all.go
monit-reviewer: **Settled**: monit-reviewer flagged that the footer hint appeared after the files section rather than after the messages table, creating a spatial disconnect. rianjs moved the hint to immediately follow the messages paging line, before the files section.

internal/output/output.go
monit-reviewer: **Settled**: monit-reviewer identified a discontinuity in truncateRunes at boundary caps 4-5 where the guard threshold is off by one. rianjs declined to fix it, noting no caller passes caps in that range and changing the guard would widen the API contract unnecessarily.

internal/cmd/search/messages.go
monit-reviewer: **Settled**: monit-reviewer flagged inconsistent footer spacing between messages.go and all.go. rianjs made both call sites print the footer with no leading blank line, directly after the paging line.

internal/cmd/search/search_test.go
monit-reviewer: **Settled**: monit-reviewer noted that the hasMessages footer guard in all.go had no positive test coverage, leaving a regression risk. rianjs added TestRunSearchAll_FooterShownWithMessages as the positive companion test.

internal/messageref/messageref.go
monit-reviewer: **Settled**: monit-reviewer flagged that a single-letter channel capture in the permalink regex could produce a misleading validation error. rianjs tightened the regex to require at least two characters, preventing the malformed input from reaching the validator.

internal/messageref/messageref.go
monit-reviewer: **Settled**: monit-reviewer suggested loosening the hardcoded \d{16} permalink timestamp regex for resilience. rianjs declined, noting Slack's format has been stable for years and the existing validate.Timestamp call already enforces the digit shape after splitting.

internal/output/output.go
monit-reviewer: **Settled**: monit-reviewer noted SearchTable always emits a header even with no rows, which could look wrong. rianjs declined to add a guard, explaining all current call sites already gate on non-empty matches, making the lonely-header path unreachable.

internal/cmd/search/search_test.go
```diff
@@ -943,3 +943,131 @@ func TestSearchMessages_WithSortOptions(t *testing.T) {
...
+		},
+	}
+	c, server := newTestClient(func(w http.ResponseWriter, _ *http.Request) {
+		_ = json.NewEncoder(w).Encode(response)
+	})
+	defer server.Close()
+	out := captureOutput(t, func() {
+		if err := runSearchAll("hi", &allOptions{count: 20, page: 1, sort: "score", sortDir: "desc"}, c); err != nil {
+			t.Fatalf("runSearchAll: %v", err)
+		}
```
monit-reviewer, line 1068: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): TestRunSearchAll_FooterShownWithMessages (and the parallel test in search_test.go:1059) uses "ts": "1.0" as the fixture timestamp. The resulting REF "C123/1.0" would be rejected by messageref.Parse because validate.Timestamp requires 10.6 digit form. The test validates footer presence but not that emitted REFs are actually usable by `messages read` — the primary workflow this PR enables. A realistic timestamp like "1234567890.123456" would exercise the full round-trip property.

internal/cmd/search/all.go
```diff
@@ -131,15 +132,16 @@ func runSearchAll(query string, opts *allOptions, c *client.Client) error {
 				Attachments: m.Attachments,
 				Files:       m.Files,
 			}, nil).Body
-			text := truncateText(body, 60)
-			ts := formatTimestamp(m.TS)
-			rows = append(rows, []string{m.Channel.Name, m.Username, ts, text})
+			when := formatTimestamp(m.TS)
```
monit-reviewer, line 135: 🔵 **Low** (harness-engineering:harness-architecture-reviewer): messageref.Ref{ChannelID: m.Channel.ID, TS: m.TS}.String() constructs a Ref directly without validating Channel.ID or TS. If Slack returns a search match with an empty Channel.ID (possible for certain DM results), the emitted REF column will contain "/1234567890.123456" — unparseable by messageref.Parse and therefore broken when a user tries to act on the footer hint. The same risk exists in search/messages.go:139. A guard on empty Channel.ID or a parse-then-format approach would surface the anomaly at display time.

internal/cmd/search/messages.go
```diff
@@ -135,15 +136,16 @@ func runSearchMessages(query string, opts *messagesOptions, c *client.Client) er
 			Attachments: m.Attachments,
 			Files:       m.Files,
 		}, nil).Body
-		text := truncateText(body, 60)
-		ts := formatTimestamp(m.TS)
-		rows = append(rows, []string{m.Channel.Name, m.Username, ts, text})
+		when := formatTimestamp(m.TS)
```
monit-reviewer, line 139: 🔵 **Low** (harness-engineering:harness-enforcement-reviewer): Same unvalidated Ref construction as all.go:135: messageref.Ref{ChannelID: m.Channel.ID, TS: m.TS}.String() emits an unparseable REF when Channel.ID is empty, which the footer then instructs the user to pass to `messages read`.

internal/messageref/messageref.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/open-cli-collective/slack-chat-api/internal/validate"
+)
+
+// conversationIDPattern accepts the full set of Slack conversation IDs used as
+// the channel portion of a message ref: public channels (C), private/group
+// channels (G), and DMs (D). validate.ChannelID rejects D, which is correct
+// for channel-only commands but too narrow for refs sourced from search
+// (which can hit DMs with --in @user).
+var conversationIDPattern = regexp.MustCompile(`^[CGD][A-Z0-9]+$`)
```
monit-reviewer, line 23: 🔵 **Low** (harness-engineering:harness-knowledge-reviewer): conversationIDPattern accepts very short IDs such as "C1" (one character after the prefix letter). Real Slack conversation IDs are always 9+ characters. Parse will accept these strings without error, but they will produce a confusing API error from Slack rather than a clear parse-time rejection. Adding a minimum-length constraint (e.g. at least 8 characters after the prefix) would catch obvious typos earlier.

monit-reviewer (approved):
## Automated PR Review



**Reviewed commit:** `331164d` | **Previous:** `beca3be` (incremental)

> **Approved** with 4 non-blocking suggestions below. Address at your discretion.

### Summary

| Reviewer | Findings |
|----------|----------|
| harness-engineering:harness-architecture-reviewer | 2 |
| harness-engineering:harness-enforcement-reviewer | 1 |
| harness-engineering:harness-knowledge-reviewer | 1 |

<details>
<summary><strong>harness-engineering:harness-architecture-reviewer (2 findings)</strong></summary>

**💡 Suggestion** - `internal/cmd/search/search_test.go:1068`
> TestRunSearchAll_FooterShownWithMessages (and the parallel test in search_test.go:1059) uses "ts": "1.0" as the fixture timestamp. The resulting REF "C123/1.0" would be rejected by messageref.Parse because validate.Timestamp requires 10.6 digit form. The test validates footer presence but not that emitted REFs are actually usable by `messages read` — the primary workflow this PR enables. A realistic timestamp like "1234567890.123456" would exercise the full round-trip property.

**💡 Suggestion** - `internal/cmd/search/all.go:135`
> messageref.Ref{ChannelID: m.Channel.ID, TS: m.TS}.String() constructs a Ref directly without validating Channel.ID or TS. If Slack returns a search match with an empty Channel.ID (possible for certain DM results), the emitted REF column will contain "/1234567890.123456" — unparseable by messageref.Parse and therefore broken when a user tries to act on the footer hint. The same risk exists in search/messages.go:139. A guard on empty Channel.ID or a parse-then-format approach would surface the anomaly at display time.

</details>

<details>
<summary><strong>harness-engineering:harness-enforcement-reviewer (1 findings)</strong></summary>

**💡 Suggestion** - `internal/cmd/search/messages.go:139`
> Same unvalidated Ref construction as all.go:135: messageref.Ref{ChannelID: m.Channel.ID, TS: m.TS}.String() emits an unparseable REF when Channel.ID is empty, which the footer then instructs the user to pass to `messages read`.

</details>

<details>
<summary><strong>harness-engineering:harness-knowledge-reviewer (1 findings)</strong></summary>

**💡 Suggestion** - `internal/messageref/messageref.go:23`
> conversationIDPattern accepts very short IDs such as "C1" (one character after the prefix letter). Real Slack conversation IDs are always 9+ characters. Parse will accept these strings without error, but they will produce a confusing API error from Slack rather than a clear parse-time rejection. Adding a minimum-length constraint (e.g. at least 8 characters after the prefix) would catch obvious typos earlier.

</details>

*5 info-level observations excluded. Run with `--verbose` to include.*

*13 PR discussion threads considered.*

---
<details>
<summary>Completed in 3m 59s | $1.12</summary>

| Field | Value |
|---|---|
| Mode | Re-review · Cycle 2 · Session resumed |
| Reviewers | hybrid-synthesis, harness-engineering:harness-architecture-reviewer, harness-engineering:harness-enforcement-reviewer, harness-engineering:harness-knowledge-reviewer, harness-engineering:harness-self-documenting-code-reviewer, security:security-code-auditor |
| Engine | claude |
| Duration | 3m 59s wall · 9m 16s compute (Reviewers: 1m 59s · Synthesis: 1m 21s) |
| Cost | $1.12 |
| Tokens | 391.7k in / 34.0k out |
| Turns | 21 |

</details>

PR #164, opened 2026-05-19
ci: build darwin releases with cgo so macOS Keychain backend is present [INT-446]

### LINKED ISSUE #163: slck macOS binaries fail closed: CGO_ENABLED=0 strips the Keychain backend [INT-446]
## Problem

Released `slck` macOS binaries fail closed on every credential operation. Root cause: `.goreleaser.yaml` sets `CGO_ENABLED=0` and the release job runs on `ubuntu-latest`. `99designs/keyring/keychain.go` is `//go:build darwin && cgo`, so with cgo disabled the macOS Keychain backend is **not compiled in** — `credstore.Open()` finds no keychain backend on darwin and fails closed. Every Mac user who upgraded since the Phase B credstore migration (#158) is broken.

Scope verified **macOS-only**: Windows (`danieljoos/wincred`) and Linux Secret Service (`godbus/dbus`) backends are pure Go and already work under `CGO_ENABLED=0`.

## Fix (pipeline only — no Go source change)

- `.goreleaser.yaml`: split builds — darwin `CGO_ENABLED=1` (per-arch `CC` overrides), linux/windows stay `CGO_ENABLED=0` static.
- `release.yml`: goreleaser job → `macos-latest` (darwin+cgo cannot cross-compile from Linux).
- Add a **release-time smoke gate**: run the freshly built darwin binary; fail the release if credstore reports no keychain backend / fail-closed.
- Fix GNU→BSD `sed` in the Homebrew-tap step (job now runs on macOS).

## Non-goals

- No runtime env credential fallback (would unwind the §1.11 keyring-only invariant).
- No cgo on linux/windows (introduces a glibc dynamic-link portability regression; those backends are pure Go).

Pilot for the cross-CLI fix tracked in Jira INT-446 (child of INT-310). Lands as `fix:` so `auto-release.yml` cuts the corrected release.

rianjs:
## Problem

Released `slck` macOS binaries fail closed on every credential op. `.goreleaser.yaml` set `CGO_ENABLED=0` and the release job ran on `ubuntu-latest`; `99designs/keyring`'s Keychain backend is `//go:build darwin && cgo`, so with cgo off it was never compiled in. On macOS `credstore` auto-selects the Keychain backend then `keyring.Open` returns `ErrNoAvailImpl` → fail-closed. Every Mac user who upgraded since the credstore migration (#158) is broken. Verified **macOS-only** (Windows wincred and Linux godbus secret-service are pure Go, fine under `CGO_ENABLED=0`).

## Fix (pipeline-only — no Go source change)

- **`.goreleaser.yaml`**: split builds — `slck-darwin` (`CGO_ENABLED=1`, fully-keyed per-arch `CC=xcrun clang -arch …` overrides for `darwin_amd64_v1` / `darwin_arm64_v8.0`) + `slck-unix-win` (`CGO_ENABLED=0` static). `nfpms.ids: [slck-unix-win]` (not the deprecated `builds`). **Both macOS arches (Intel + Apple Silicon) still produced** — the split is by GOOS, not arch.
- **`release.yml`**: goreleaser job → pinned `macos-15` (cgo+darwin can't cross-compile from Linux). Restructured to **verify before publish**: `goreleaser check` → `release --snapshot` (no publish) → pre-publish gate → real `release --clean --release-notes`. The gate parses `dist/artifacts.json` and asserts: arm64 binary functionally reports `backend=keychain backend_source=auto credential_ref=slack-chat-api/default` (isolated HOME/XDG, `-u SLACK_CHAT_API_KEYRING_BACKEND`); amd64 binary links `Security.framework` (necessary cgo signal for the slice the runner can't execute); both darwin archives present exactly once. A backend-less darwin binary can no longer be published silently.
- Hardened the `update-homebrew` checksum step (`set -euo pipefail` + non-empty SHA assertions) so a renamed/missing archive can't publish a broken cask.

## Non-goals

- No cgo on linux/windows (glibc portability regression, zero benefit — pure-Go backends).
- No runtime env credential fallback (would unwind the §1.11 keyring-only invariant).
- The pre-existing `workflow_dispatch` `TAG: github.ref_name || inputs.tag` bug is out of scope (only affects the dispatch path; this fix uses the tag-push path) — separate ticket.

## Validation

`goreleaser check` passes; both cgo cross-builds succeed locally (arm64 native + amd64 cross, correct Mach-O archs, amd64 links Security.framework); the gate's functional assertion reproduced on real hardware (`backend=keychain/auto`).

## Release mechanics

`auto-release.yml` Gate-1 is a path filter (`**.go|go.mod|go.sum`); a pipeline-only diff matches none, so this is a `ci:` commit that does not auto-release. After merge the corrected release is cut by a deliberate annotated tag push (`release.yml` `on: push: tags: v*`).

Pilot for the cross-CLI CGO regression (Jira INT-446, child of INT-310); verified on a non-dev Mac before any fan-out to nrq/gro/atlassian.

Closes #163

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,10 @@ permissions:
 
 jobs:
   goreleaser:
-    runs-on: ubuntu-latest
+    # INT-446: darwin must build with cgo (Keychain backend). cgo+darwin
+    # cannot cross-compile from Linux, so this job runs on macOS. Pinned
+    # image (not the moving macos-latest label) for a reproducible release.
+    runs-on: macos-15
     env:
       TAG: ${{ github.ref_name || inputs.tag }}
     steps:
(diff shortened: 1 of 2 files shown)


rianjs:
**Findings**
- None.

The PR adheres to the converged INT-446 plan. The diff is scoped to `.goreleaser.yaml` and `release.yml`; no unrelated code or product behavior crept in. GoReleaser split is correct: darwin is CGO-on with fully keyed overrides and quoted `CC`, linux/windows remain CGO-off, `nfpms.ids` targets the non-darwin build, and archives remain unfiltered with disjoint GOOS sets.

The pre-publish gate is sound: snapshot build happens before publish, artifact lookup uses `artifacts.json` by type/OS/arch, arm64 proves `keychain/auto` through the shipped binary under isolated HOME/XDG, amd64 gets the necessary `Security.framework` check, and release notes are verified with `GH_TOKEN`. Bash is macOS-safe: no Bash 4 constructs, `set -euo pipefail` is present, heredoc and `$RUNNER_TEMP` are quoted, and Homebrew checksum parsing now fails before writing empty SHAs. GoReleaser Action supports `version: "~> v2"` and `install-only` per its docs.

PR #186, opened 2026-06-04
chore: align darwin codesign hook to the fail-loud form

rianjs:
slck carried the original *simple* darwin codesign hook; the other five family CLIs (cfl/jtk/gro/nrq/cr) ship the hardened *fail-loud* form — where `CODESIGN_DARWIN_SCRIPT` set-but-missing or non-executable **errors** instead of silently skipping. This makes slck's hook byte-identical to theirs so the convention (cli-common `distribution.md §2A`) is uniform across all six repos.

- YAML-only; `.goreleaser.yaml` is not a gated release path → no release is cut.
- `goreleaser check -f .goreleaser.yaml` passes.
- Verified byte-identical to the cfl hook.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -48,7 +48,7 @@ builds:
     # local builds → signing skipped.
     hooks:
       post:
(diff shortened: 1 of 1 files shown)


rianjs:
**Codex architect review** (continuity session from the signing initiative)

The hook is now byte-identical to the refined fail-loud convention. Unset env still no-ops for local builds; set-but-missing/non-executable fails before an unsigned intended-to-sign build can proceed. `goreleaser check` passes.

STATUS: blockers=0 majors=0 minors=0 nits=0

